Issue #55, opened 2016-08-03
Support geometry simplification


ARolek:

gdey:
Clipping takes care of this as well.

PR #75, opened 2016-08-20
Initial Simplification code.

gdey:
This code works and simplify the points as one zooms out. The performance of the code is slowed down as
it is having to loop through all the points one more time. May be able to optimize this later. Also, there is a possible bug with features that become too small to show. We return nil, but not sure how the encoding layer is going to deal with this. It should ignore this, but this could mean that a feature totally disappears. In which case the layer it's self should not be rendered. Still have to test this.

Also, missing are test cases.

This is the pull request for issue #55. When it is ready I will comment.

### DIFF
--- a/basic/line.go
+++ b/basic/line.go
@@ -12,22 +12,22 @@ func (Line) basicType()     {}
 func (Line) String() string { return "Line" }
 
 // NewLine creates a line given pairs for floats.
-func NewLine(pointPairs ...float64) *Line {
-	line := new(Line)
+func NewLine(pointPairs ...float64) Line {
+	var line Line
 	if (len(pointPairs) % 2) != 0 {
 		panic("NewLine requires pair of points.")
 	}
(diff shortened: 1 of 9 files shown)


basic/line.go
```diff
@@ -14,7 +14,7 @@ func (Line) String() string { return "Line" }
 // NewLine creates a line given pairs for floats.
 func NewLine(pointPairs ...float64) *Line {
 	line := new(Line)
-	if (2 % len(pointPairs)) != 0 {
+	if (len(pointPairs) % 2) != 0 {
 		panic("NewLine requires pair of points.")
```
ARolek: Do we want to set a precedence that a panic is ok in this situation? I understand we avoid returning an error which is nice...

gdey: That's a good question. This panic has saved me. And the `2 % len(poinPairs)` is bug that has slipped by us. Could change it to return an error, and I am inclined to lean that way.

PR #93, opened 2017-01-11
Tegola on Docker

waybarrios:
This is dockerfile which it is possible  to compile binary by itself, and running for everyone who wants to use it. It is public on Docker Hub: terranodo/tegola image. We need to discuss future work to use scratch instead of golang image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM golang
+MAINTAINER terranodo
+RUN mkdir -p /tegola/
+WORKDIR /tegola/
+COPY . /tegola
+
+#Requirements to compile 
+RUN mkdir -p /usr/local/go/src/github.com/terranodo/
+RUN cp -r /tegola/ /usr/local/go/src/github.com/terranodo/
+RUN mkdir -p /usr/local/go/src/github.com/BurntSushi/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM golang
```
ARolek: I think we should be explicit about the version here. We currently require go 1.7, and since there a few patch versions we should go with `FROM golang:1.7.4`.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+MAINTAINER terranodo
+RUN mkdir -p /tegola/
+WORKDIR /tegola/
+COPY . /tegola
+
+#Requirements to compile 
+RUN mkdir -p /usr/local/go/src/github.com/terranodo/
+RUN cp -r /tegola/ /usr/local/go/src/github.com/terranodo/
+RUN mkdir -p /usr/local/go/src/github.com/BurntSushi/
```
ARolek: This step should not be necessary as we vendor the BurntSushi package. Are you having problems compiling without it? If so, I need to look into the vendor directory.

JivanAmara:
Closing in anticipation of #244 .

PR #108, opened 2017-04-11
Issue 81

ARolek:


config/config.go
```diff
@@ -15,46 +15,107 @@ import (
...
 type Config struct {
 	// LocationName is the file name or http server that the config was read from. If this is "", it means that the location was unknown. This is the case if the Prase() function is used
 	// directly.
 	LocationName string
-	Webserver    struct {
-		Port      string
-		LogFile   string `toml:"log_file"`
-		LogFormat string `toml:"log_format"`
-	}
+	Webserver    Webserver `webserver`
```
gdey: Shouldn't the tag be `toml:"webserver"`

config/config.go
```diff
@@ -15,46 +15,107 @@ import (
...
+				if val.MinZoom <= l.MaxZoom && l.MinZoom <= val.MaxZoom {
+					return ErrLayerCollision{
+						ProviderLayer1: val.ProviderLayer,
+						ProviderLayer2: l.ProviderLayer,
+					}
+				}
+			} else {
+				//	add the MapLayer to our map
+				layerNames[plParts[1]] = l
+			}
```
gdey: I would rewrite the if statement to look like this:
```go
// Check if we already have this layer
if val, ok := layerNames[plParts[1]]; ok {
    // we have a hit, check for zoom range overlap.
    if val.MinZoom < l.MaxZoom && l.MinZoom <= val.MaxZoom {
        return ErrLayerCollision {
                  ProviderLayer1 : val.ProviderLayer,
                  ProviderLayer2: l.ProviderLayer,
        }
    }
    continue
}

// add the MapLayer to our map
layerNames[plParts[1]] = l
```

PR #121, opened 2017-05-11
Plumbed the context struct through the tile generation. [issue-111]

### LINKED ISSUE #111: Implement request context
Incoming tile requests have no way of being canceled right now. Canceling a request is something that happens often when an user is zooming and panning around a map. Without a way to cancel a request, tegola will always do the work necessary to generate a tile even though it's no longer needed. 

Implementing context.Context will allow for the web server to listen for request cancelation and communicate that to additional internal packages.

gdey:
Need to plumb the context for cancellation throught tile generation.
I did not plumb it into the scaling, or clipping functions as we can
cancel between the operations, that way we don't incure extra check
when comparing lot's of points.

If we need to plumb it futher down later, we will be able to.

Fixes #111


mvt/provider.go
```diff
@@ -1,11 +1,20 @@
...
-//Provider is the mechinism by which the system talks to different data providers.
+	"github.com/terranodo/tegola"
+)
+
+var ErrCanceled = context.Canceled
+
+//Provider is the mechanism by which the system talks to different data providers.
 type Provider interface {
 	// MVTLayer returns a layer object based
 	MVTLayer(layerName string, tile tegola.Tile, tags map[string]interface{}) (*Layer, error)
```
gdey: We may want to remove this definition in the Interface. I left it in for backward compatibility, but am now thinking this is a mistake. I think it makes more sense to just have people use the MVTLayerWithContext.

ARolek: I partially agree. I think we should keep the function named MVTLayer but add the context to the function signature. No need to name it MVTLayerWithContext unless we also support without context. I think this naming convention should be applied throughout.

Issue #105, opened 2017-03-14
Fix spelling mistakes reported by go report.


gdey:
Go Report is currently reporting several [spelling errors](https://goreportcard.com/report/github.com/terranodo/tegola#misspell), fix these.

* [tegola/basic/line.go](https://github.com/terranodo/tegola/blob/master/basic/line.go)
  *   [Line 6](https://github.com/terranodo/tegola/blob/master/basic/line.go#6): warning: "interect" is a misspelling of "interacted" (misspell)
* [tegola/basic/polygon.go](https://github.com/terranodo/tegola/blob/master/basic/polygon.go)
  * [Line 29](https://github.com/terranodo/tegola/blob/master/basic/polygon.go#29): warning: "retuns" is a misspelling of "returns" (misspell)
* [tegola/mvt/feature.go](https://github.com/terranodo/tegola/blob/master/mvt/feature.go)
  * [Line 133](https://github.com/terranodo/tegola/blob/master/mvt/feature.go#133): warning: "Untill" is a misspelling of "Until" (misspell)
* [tegola/mvt/feature_test.go](https://github.com/terranodo/tegola/blob/master/mvt/feature_test.go)
  * [Line 193](https://github.com/terranodo/tegola/blob/master/mvt/feature_test.go#193): warning: "TOOD" is a misspelling of "TODO" (misspell)
* [tegola/mvt/tile.go](https://github.com/terranodo/tegola/blob/master/mvt/tile.go#)
  * [Line 37](https://github.com/terranodo/tegola/blob/master/mvt/tile.go#37): warning: "everthing" is a misspelling of "everything" (misspell)
* [tegola/provider/postgis/postgis.go](https://github.com/terranodo/tegola/blob/master/postgis/postgis.go#)
  * [Line 27](https://github.com/terranodo/tegola/blob/master/postgis/postgis.go#27): warning: "soemthing" is a misspelling of "something" (misspell)
  * [Line 37](https://github.com/terranodo/tegola/blob/master/postgis/postgis.go#37): warning: "corrosponding" is a misspelling of "corresponding" (misspell)
  * [Line 411](https://github.com/terranodo/tegola/blob/master/postgis/postgis.go#411): warning: "responsiblity" is a misspelling of "responsibility" (misspell)
tegola/server/handle_map_capabilities.go
  * [Line 87](https://github.com/terranodo/tegola/blob/master/postgis/postgis.go#87): warning: "determing" is a misspelling of "determining" (misspell)

PR #151, opened 2017-10-12
Fix misspellings reported in #105.

erictheise:
Additionally, several misspellings of 'geometery' -> 'geometry'.


ARolek:
@erictheise thank you! we will probably need to do this again when v0.4.0 is cut. big release coming!

Issue #163, opened 2017-10-30
504 returned when attempting to retrieve a tile at negative zoom


jamesyarrington:
**To Reproduce:**
Send a GET to:
```
osm.tegola.io/maps/osm/-1/0/0.pbf
```
**Expected Response:**
A `400 Bad Request`should be returned, along with a message stating that negative zoom levels are not allowed.

**Actual Response:**
A `504 Gateway Time-out` is returned, with no other message.

ARolek:
@jamesyarrington thanks for the report.

ARolek:
Fixed in v0.4.0 will be in the final version release.

PR #168, opened 2017-11-01
[issue-163] If the z value is neg return a bad request.

gdey:
Zoom value is now check to see if it's greater then or equal to zero.
If it isn't log that we got a bad value and return a 400 error.


gdey:
This is regards to issue #163

PR #201, opened 2017-11-29
Issue199 host name fixup

JivanAmara:
Changes to server.hostName() to:

1. return <host>:<port> instead of <host>
2. accept ":<port>" or "<port>" from config

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -26,8 +26,6 @@ var (
 	Port string
 	//	reference to the version of atlas to work with
 	Atlas *atlas.Atlas
-	//	cache interface to use
-	Cache cache.Interface
 )
 
 //	Start starts the tile server binding to the provided port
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -58,17 +61,45 @@ func Start(port string) {
...
+		portNumber = portNumber[1:]
+	}
+
+	if portNumber == "none" {
+		// Don't add a port to the host.
+	} else if portNumber != "" {
+		retHost += ":" + portNumber
+	} else if requestPort != "" {
+		retHost += ":" + requestPort
 	}
```
gdey: I would rewrite this `if else` statement to:
```
if portNumber != "" && portNumber != "none" {
      return retHost + ":" + portNumber
}
if requestPort != "" {
     return retHost + ":" + portNumber
}
return retHost
```

JivanAmara: Don't see any benefit to this change, but will make.

JivanAmara:
Will re-issue pull request without config port changes.

Issue #147, opened 2017-09-22
Redo logging output.


sacontreras:
1. tegola - at least in 0.4.0 branch - seems to send output meant for stdout to stderr - e.g. loading config notification and successful server start messages are output to stderr, not stdout
2. in fact all output *appears* to be sent to stderr as I have yet to see any output at all sent to stdout

ARolek:
This will require reworking how we're handling logging in tegola. Currently we're mainly using `log.Println(...)` which defaults to `os.Stderr`. I'm in favor of cleaning this up and making log levels configurable. @goinggo has an article on this topic we can use for inspiration: [Using The Log Package In Go](https://www.goinggo.net/2013/11/using-log-package-in-go.html) which was turned into the [tracelog](https://github.com/goinggo/tracelog) package.

gdey:
We need to redo our logging output.
For this issue.

1. Vendor in the [tracelog](https://github.com/goinggo/tracelog) package.
2. Added loggers for:
- [ ] Trace
- [ ] Info
- [ ] Error
- [ ] Warning
3. Update log statements to the appropriate logger.

ARolek:
**Logger Proxy API**
- `log.Fatal(string, vals... interface{})` -
  - prevents the program from continuing
  - i.e. can't allocate additional memory
- `log.Error(string, vals... interface{})`
  - non-fatal, but prevents valid execution
  - i.e. can't connect to a database, complete a function call, open file, invalid format
- `log.Warn(string, vals... interface{})`
  - looks unusual, but does not clearly prevent execution
- `log.Info(string, vals... interface{})`
  - i.e. server request logs
- `log.Debug(string, vals... interface{})`
  - high level info for a developer. more than what a sysadmin would typically want
- `log.Trace(string, vals... interface{})`
  - excruciating detail. first cut will try to non compile the statements out. if we find we're still dealing with performance issues, then we will explore turning `log.Trace(...)` into a noop.

**Aux Functions**

Default Log Level is Info.
- SetLogLevel(log.Level)

```go
// tegola/internal/log
package log

var (
   lock syc.Mutex
  // in order 
  IsError bool
  IsWarn bool
  IsInfo bool
  IsDebug bool
  IsTrace bool
)
/* example usage
if log.IsDebug {

}
*/

```
**Logger requirements**
- [ ] if a log is called that is above the set log level, then the function should not do work (just return)
- [ ] log level should be set via the config file
- [ ] log level should be set via an ENV var
[...]

PR #205, opened 2017-12-02
Log proxy w/ std logger implementation & tests.

JivanAmara:
See issue #147 for details.

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -45,6 +45,10 @@ var (
 	IsTrace bool
 )
 
+func SetOutput(w io.Writer) {
+	logger.SetOutput(w)
+}
+
 func SetLogLevel(lvl Level) {
 	lock.Lock()
 	IsTrace = false
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -1,0 +1,126 @@
+package log
+
+import (
+	"io"
+	"sync"
+)
+
+type Logger interface {
+	// These all take args the same as calls to fmt.Printf()
+	Fatal(string, ...interface{})
+	Error(string, ...interface{})
+	Warn(string, ...interface{})
+	Info(string, ...interface{})
+	Debug(string, ...interface{})
+	Trace(string, ...interface{})
+	SetOutput(io.Writer)
+}
+
+func init() {
+	defaultLogger = &stdLogger
+}
+
+// Allows the ordering of severity to be checked
+const (
+	FATAL = 60
+	ERROR = 50
+	WARN  = 40
+	INFO  = 30
+	DEBUG = 20
+	TRACE = 10
+)
+
+var (
+	defaultLogger Logger
+	logLevel      int = INFO
+	logLock       sync.Mutex
+	IsError       bool = true
+	IsWarn        bool = true
+	IsInfo        bool = true
+	IsDebug       bool = false
+	IsTrace       bool = false
+)
+
+func SetLogLevel(level int) {
+	logLock.Lock()
+	switch {
+	case level <= TRACE:
+		logLevel = TRACE
+		IsError = true
+		IsWarn = true
+		IsInfo = true
+		IsDebug = true
+		IsTrace = true
+	case level <= DEBUG:
+		logLevel = DEBUG
+		IsError = true
+		IsWarn = true
+		IsInfo = true
+		IsDebug = true
+		IsTrace = false
+	case level <= INFO:
+		logLevel = INFO
+		IsError = true
+		IsWarn = true
+		IsInfo = true
+		IsDebug = false
+		IsTrace = false
+	case level <= WARN:
+		logLevel = WARN
+		IsError = true
+		IsWarn = true
+		IsInfo = false
+		IsDebug = false
+		IsTrace = false
+	case level <= ERROR:
+		logLevel = ERROR
+		IsError = true
+		IsWarn = false
+		IsInfo = false
+		IsDebug = false
+		IsTrace = false
+	default:
+		logLevel = FATAL
+		IsError = false
+		IsWarn = false
+		IsInfo = false
+		IsDebug = false
+		IsTrace = false
+	}
+	logLock.Unlock()
+}
+
+// Output format should be: "timestamp•LOG_LEVEL•filename.go•linenumber•output"
+func Fatal(format string, args ...interface{}) {
+	defaultLogger.Fatal(format, args...)
+}
+
+func Error(format string, args ...interface{}) {
+	if IsError {
+		defaultLogger.Error(format, args...)
+	}
+}
+
+func Warn(format string, args ...interface{}) {
+	if IsWarn {
+		defaultLogger.Warn(format, args...)
+	}
+}
+
+func Info(format string, args ...interface{}) {
+	if IsInfo {
+		defaultLogger.Info(format, args...)
+	}
+}
+
+func Debug(format string, args ...interface{}) {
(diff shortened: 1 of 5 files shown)


internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
+package log
+
+import (
+	"fmt"
+	"io"
+	goLog "log"
```
gdey: No need to do this. Just leave it as log.

internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+	"runtime"
+)
+
+var stdLogger StdLogger
+
+type StdLogger struct {
```
gdey: Just call it `Standard`, and not `StdLogger` or just 'Std`

internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"io"
+	goLog "log"
+	"os"
+	"runtime"
+)
+
+var stdLogger StdLogger
```
gdey: Just call this `standard` or `std`.

internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
...
+
+type StdLogger struct {
+	Logger
```
gdey: Why are you embedding an interface? Looks like this should just be an empty struct{}:
```
type Standard struct{}
```

internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func SetOutput(w io.Writer) {
+	goLog.SetOutput(w)
+}
+
+func Output(level string, format string, args ...interface{}) {
```
gdey: This should be part of Standard and private:
```
func (_ Standard) output(…){…}
```
Or just private.

internal/log/stdLogger.go
```diff
@@ -0,0 +1,51 @@
...
+	logMsg := fmt.Sprintf(format, args...)
+	_, file, line, _ := runtime.Caller(3)
+	// timestamp will be provided by goLog
+	goLog.Printf("•%v•%v•%v•%v", level, file, line, logMsg)
+}
+
+func (l *StdLogger) Fatal(format string, args ...interface{}) {
```
gdey: No need to the pointer here, you are not making use it. Just have it be of the empty struct.
```
func (_ Standard)Fatal(…)
```
This of course applies to the rest of these functions.

internal/log/log_test.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"bytes"
+	"fmt"
+	"regexp"
+	"testing"
+)
+
+func TestSetLevel(t *testing.T) {
+	// Check default level is INFO & Is* flags are set appropriately
+	if !(logLevel == INFO && IsError == true && IsWarn == true && IsInfo == true && IsDebug == false &&
```
gdey: For boolean value no need to do the `var == true` just call the `var`. i.e:
```
if level != TRACE || !IsError || !IsWarn || !IsInfo || !IsDebug || !IsTrace {
      t.Errorf("TRACE level is not set as expected.")
}
```

I would go as far as making a helper function:
```
func levelChecksOut(lvl Level)(ok bool) {
    if level != lvl  {
        return false
   }
   switch lvl {

      case TRACE:
           if !IsTRACE {
                 return false
           }
           fallthrough
      case DEBUG:
           if !IsDEBUG {
                return false
          }
           fallthrough
      default: 
        case INFO:
          if !IsINFO {
                 return false
          }
           fallthrough
       case WARN:
           if !IsWARN {
                 return false
           }
           fallthrough
         case ERROR:
            if !IsERROR {
                  return false
           }
           fallthrough
         case FATAL:
          if !IsFATAL {
                return false
          }
           fallthrough
       }
     return true
}
```
The in the test function:
```
if levelChecksOut(TRACE) {
      t.Errorf("TRACE level is not set as expected.")
}
```

internal/log/stdLogger.go
JivanAmara: Inside a "log" package, I think this makes references more clear.

internal/log/stdLogger.go
JivanAmara: While it only functionally only needs to be an empty struct, this makes it clear to someone viewing the file that the struct is intended to implement the interface.

internal/log/log.go
```diff
@@ -0,0 +1,126 @@
...
+	Error(string, ...interface{})
+	Warn(string, ...interface{})
+	Info(string, ...interface{})
+	Debug(string, ...interface{})
+	Trace(string, ...interface{})
+	SetOutput(io.Writer)
+}
+
+func init() {
+	defaultLogger = &stdLogger
```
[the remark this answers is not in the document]
JivanAmara: default is a reserved word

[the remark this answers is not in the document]
gdey: Oh right, Forgot about that. Always bytes me. Ignore this comment.

internal/log/stdLogger.go
gdey: I would make `Output` private; as we don't want anyone outside of the package directly call this function.

internal/log/log.go
```diff
@@ -17,110 +17,96 @@ type Logger interface {
...
 func Fatal(format string, args ...interface{}) {
-	defaultLogger.Fatal(format, args...)
+	logger.Fatal(format, args...)
 }
 
 func Error(format string, args ...interface{}) {
 	if IsError {
-		defaultLogger.Error(format, args...)
+		logger.Error(format, args...)
 	}
```
gdey: Sorry, did not notice this before. Redo these if statements as such:
```
func Error(format string, args ...interface{}){
    if !IsError {
          return
    }
    logger.Error(format,args...)
}
```
This is more idiomatic.  It, also, reduces the indention level for the important statements.

internal/log/log.go
```diff
@@ -17,110 +17,96 @@ type Logger interface {
...
+	IsDebug = false
+	IsInfo = false
+	IsWarn = false
+	IsError = false
+	IsTrace = false
+	if lvl != TRACE && lvl != DEBUG && lvl != INFO && lvl != WARN && lvl != ERROR && lvl != FATAL {
+		lvl = INFO
+	}
+	level = lvl
 	switch {
```
gdey: Why not just turn the `switch` to:
```
switch level {
   case TRACE:
     …
  case DEBUG:
   …
}
```
I think that reads better.
Since with the fallthrough, understands it's going to the include the next set of cases.

internal/log/log.go
JivanAmara: Seriously?

JivanAmara: Seems like making the changes will be faster than trying to discuss.  Will do.

internal/log/stdLogger.go
gdey: Sorry, somehow I missed this comment when I went through. This is not what you are doing here. What you are doing here is actually declaring a public Field called Logger. This field can be assigned to and called. If it's called before being given a non-nil value it will panic. None of the functions on the non-pointer version of the Struct will call any of the log functions, but the pointer version of the struct will attempt to call functions on the Logger field. 

This is not what we want. This should just be an empty struct.

gdey:
LGTM (Looks good to me.)

JivanAmara:
Got the go-ahead from @ARolek on slack.

PR #212, opened 2017-12-06
Issue199 host name fixup

JivanAmara:
Adds port to output of server.hostName()


server/handle_capabilities_test.go
```diff
@@ -120,12 +125,169 @@ func TestHandleCapabilities(t *testing.T) {
...
+			uriPattern: "/capabilities",
+			reqMethod:  "GET",
+			expected: server.Capabilities{
+				Version: serverVersion,
+				Maps: []server.CapabilitiesMap{
+					{
+						Name:         "test-map",
+						Attribution:  "test attribution",
+						Center:       [3]float64{1.0, 2.0, 3.0},
+						Bounds:       [4]float64{-180.0, -85.0511, 180.0, 85.0511},
```
gdey: Might be good to use this package variable:
https://github.com/terranodo/tegola/blob/v0.5.0/projections.go#L10

JivanAmara: Would certainly make it cleaner to read.

PR #213, opened 2017-12-08
Issue 211

ARolek:


cache/s3cache/s3cache.go
```diff
@@ -0,0 +1,242 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			default:
+				return aerr
+			}
+		}
+		return err
+	}
+
+	return nil
```
gdey: What is happening here is hidden. I think it would be easier to understand if it were:
```
return err
```

You are not doing anything special with the error here. The extra complication for the error transform and error handling is wasted.

cache/s3cache/s3cache.go
```diff
@@ -0,0 +1,242 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			default:
+				return err
+			}
+		}
+		return err
+	}
+
+	return nil
```
gdey: No need for the complicated error handling here. You are not doing anything with the error; so just return it.
```
return err
```

cache/s3cache/s3cache_test.go
```diff
@@ -0,0 +1,202 @@
...
+			config: map[string]interface{}{
+				"bucket": "tegola-test-data",
+				"region": "us-west-1",
+			},
+			key: cache.Key{
+				Z: 0,
+				X: 1,
+				Y: 2,
+			},
+			expected: []byte("\x53\x69\x6c\x61\x73"),
```
gdey: FYI: This can, also, be written as:
```
expected: []byte{0x53,0x69,0x6c,0x61,0x73}
```

cache/s3cache/s3cache.go
ARolek: good point. I was leaving it stubbed out in case we discover an error to capture, but there is no point in being prepared for something that will likely never happen.

ARolek:
Just pushed updates per your suggestions.

cache/s3cache/README.md
```diff
@@ -0,0 +1,27 @@
...
+[cache]
+type="s3"   
+bucket="tegola-test-data"
+```
+
+## Properties
+The s3cache config supports the following config properties:
+
+- `bucket` (string): [Required] the name of the S3 bucket to use.
+- `region` (string): [Optoinal] the region the bucket is in. Defaults to 'us-east-1'
```
JivanAmara: The only issue I see is pretty minor; "Optional" miss-spelled a few times in the README.

ARolek: Good catch! I will push up a fix in just a bit. Looks like this is ready to be merged. Thanks for the review.

Issue #176, opened 2017-11-09
/capabilities endpoints should include no-cache headers


ARolek:
Since the endpoints are generated and include protocols (http, https) to various resources, these endpoints should not cache. As it stands, if the resource is requested via http and cached via a CDN, subsequent requests over https will include URLS with http as their protocol.

PR #216, opened 2017-12-12
added No-Cache headers to server endpoints which should not cache. cl…

ARolek:
…oses #176


JivanAmara:
Changes look fine, should they also be made to server/handle_map_zxy.go, server/middleware_tile_cache.go, server/handle_map_layer_zxy.go?

ARolek:
@JivanAmara nope. We want those to cache at the CDN layer, or allow a cache proxy to take control. These are the only endpoints we want to explicitly not cache.

JivanAmara:
Ok, LGTM

PR #220, opened 2017-12-13
Add testing helpers in 'internal/assert'.

JivanAmara:
Will be using these in place of testify versions on gpkg provider branch.

### DIFF
--- a/internal/assert/assertions.go
+++ b/internal/assert/assertions.go
@@ -1,0 +1,130 @@
+package assert
+
+// Testing check shortcuts
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// Example Usage:
+// assert.Equal(t, rightAnswer, testAnswer)
+// assert.Equal(t, rightAnswer, testAnswer, "Append this to failure message")
+// assert.Equal(t, rightAnswer, testAnswer, "More usefully, append this: %v", someImportantValue)
+
+// Log message and fail if actualValue doesn't match expectedValue
+// *Note that DeepEqual will check the value pointed to by two pointers, ie don't use for checking
+//	pointer equality.
+func Equal(t *testing.T, expectedValue interface{}, actualValue interface{}, cm ...interface{}) bool {
+	// 'cm', the custom message arguments are optional, and if provided will be appended to a fail
+	//	message if generated.
+	// If cm[0] is a string, the cm arguments will be treated like fmt.Printf() treats its params.
+	// If cm[0] is not a string, the cm arguments will be treated like fmt.Print().
+	if !reflect.DeepEqual(expectedValue, actualValue) {
+		msg := fmt.Sprintf("%v != %v (expected)", actualValue, expectedValue)
+		if len(cm) > 0 {
+			switch cm[0].(type) {
+			case string:
+				msg = msg + " - " + fmt.Sprintf(cm[0].(string), cm[1:]...)
+			default:
+				msg = msg + " - " + fmt.Sprint(cm...)
+			}
+		}
+		return Fail(t, msg)
+	}
+	return true
+}
+
+// Log message and fail if testValue is not true.
+func True(t *testing.T, testValue bool, cm ...interface{}) bool {
+	// 'cm', the custom message arguments are optional.
+	// If cm[0] is a string, the cm arguments will be treated like fmt.Printf() treats its params.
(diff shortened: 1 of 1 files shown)


internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// Log message and fail if actualValue doesn't match expectedValue
+// *Note that DeepEqual will check the value pointed to by two pointers, ie don't use for checking
+//	pointer equality.
+func Equal(t *testing.T, expectedValue interface{}, actualValue interface{}, cm ...interface{}) bool {
+	if !reflect.DeepEqual(expectedValue, actualValue) {
+		msg := fmt.Sprintf("%v != %v (expected)", actualValue, expectedValue)
+		if len(cm) > 0 {
+			msg = msg + " - " + fmt.Sprintf(cm[0].(string), cm[1:]...)
```
gdey: You need to check if cm[0] is actually a string. The format of Equal is not idiomatic. One would think Equal works like Println. And not send in a format string.

internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+	"reflect"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+// Log message and fail if actualValue doesn't match expectedValue
+// *Note that DeepEqual will check the value pointed to by two pointers, ie don't use for checking
+//	pointer equality.
+func Equal(t *testing.T, expectedValue interface{}, actualValue interface{}, cm ...interface{}) bool {
```
ARolek: Can you provide an example of how `cm` should be used? It's not clear from the method signature.

internal/assert/assertions.go
ARolek: if the first value of `cm` is not a string, this will panic. you need to check the assertion or maybe change the function signature to accept a string then variadic interface{} params.

internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+	return true
+}
+
+// Fail performs t.Errorf(), but modifies the message with the file and line of the caller.
+func Fail(t *testing.T, failureMessage string) bool {
+	//	content := [][2]string{
+	//		{"Error Trace", strings.Join(CallerInfo(), "\n\r\t\t\t")},
+	//		{"Error", failureMessage},
+	//	}
+	_, file, line, _ := runtime.Caller(2)
```
ARolek: is an error being ignored here? Maybe throw a panic on error if you don't want to return it in this instance?

internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+			msg = msg + " - " + fmt.Sprintf(cm[0].(string), cm[1:]...)
+		}
+		return Fail(t, msg)
+	}
+	return true
+}
+
+// Fail performs t.Errorf(), but modifies the message with the file and line of the caller.
+func Fail(t *testing.T, failureMessage string) bool {
+	//	content := [][2]string{
```
ARolek: can this code be removed?

internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+	//		{"Error Trace", strings.Join(CallerInfo(), "\n\r\t\t\t")},
+	//		{"Error", failureMessage},
+	//	}
+	_, file, line, _ := runtime.Caller(2)
+	fileParts := strings.Split(file, "/")
+	file = fileParts[len(fileParts)-1]
```
ARolek: Can `fileparts` ever be 0? if so, this will panic.

internal/assert/assertions.go
```diff
@@ -0,0 +1,92 @@
...
+	// The "\r" here is what eliminates the default file:line portion of the message.
+	t.Errorf("\r\t%s:%v: %s", file, line, failureMessage)
+
+	return false
+}
+
+func isNil(object interface{}) bool {
```
ARolek: maybe name `object` to `val` since an interface{} != struct

internal/assert/assertions.go
JivanAmara: With a check after runtime.Caller(), no, fileparts won't ever be of length 0.

JivanAmara:
I added some usage examples and extended the custom message args (cm) to take either the args to a Printf() or Print() -like function.  The primary reason for the variable parameter is so people can happily ignore adding a custom message when it's not needed (the majority of cases).

I added a check & panic for the possibility of runtime.Caller() failing.

ARolek:

PR #222, opened 2017-12-14
Logging interface now has formatting and raw log methods, i.e. Errorf() / Error()

JivanAmara:
Go code regularly uses calls like
```
f, err := os.Open("filename.ext")
if err != nil {
    log.Fatal(err)
}
```

This updates the logger to accept an error var in addition to a string.

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -5,6 +5,8 @@ import (
 	"sync"
 )
 
+var TimestampRegex string = `\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}` // Ex: "2006-01-02 15:04:05"
+
 type Interface interface {
 	// These all take args the same as calls to fmt.Printf()
 	Fatal(string, ...interface{})
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -1,6 +1,7 @@
 package log
 
 import (
+	"fmt"
 	"io"
 	"sync"
 )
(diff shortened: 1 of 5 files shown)


internal/log/log.go
```diff
@@ -81,41 +81,83 @@ func SetLogLevel(lvl Level) {
 }
 
 // Output format should be: "timestamp•LOG_LEVEL•filename.go•linenumber•output"
-func Fatal(format string, args ...interface{}) {
-	logger.Fatal(format, args...)
+func Fatal(msg interface{}, args ...interface{}) {
```
ARolek: I'm wondering if we should setup a second set of methods like `Fatalf`, Errorf, Warnf, etc. to match with the convention of having a message format string like the stdlib `log` package uses.

internal/log/log.go
```diff
@@ -81,41 +81,83 @@ func SetLogLevel(lvl Level) {
...
+	var msgString string
+	switch m := msg.(type) {
+	case string:
+		msgString = m
+	case error:
```
ARolek: I have not implemented this package yet, but are you finding an use case for passing in error types as format strings? If it's not a common use case, I would prefer that we keep with the string type over the interface type so we get compile time errors rather than runtime errors. If this is an use case, then we should add a `default` statement for the case when the type is not `string` or `error`. I'm not quite sure how to handle it though. I don't think it's a good idea to panic here, but returning an error will also require all logging to then capture errors.

internal/log/log.go
gdey: I agree here. We should have two set's of functions, following the way the stdlib has set things up. The `f` variant should take a format string, and the non-f variant should not just take a bunch of interfaces and just pass them on with doing anything special to them.

JivanAmara: Are you guys sure?  I think it's a good idea that it matches the standard log interface and suggested it during the design meeting, but you didn't want it then.

internal/log/log.go
JivanAmara: The standard log package accepts an error or a string and both are currently being used.  I made this change to allow the package to replace existing logging without having to update calls to use an "log.Error(err.Error())" instead of simply"log.Error(err)".

internal/log/log.go
gdey: I think there may have been a misunderstanding; during that meeting. I do not recall not wanting to follow the standard log interface. If that was the case I apologize, we do want to follow the go-developers standards as much as possible — as long as it makes sense.

Issue #107, opened 2017-04-10
configurable tile buffer


ARolek:
One of the issues being reported during geometry clipping (#56) is a rendering fragment due to the lack of tile buffer. Vector tiles use a "hack" where the tile buffer extends beyond the tile extent. This allows for a smooth transition between tiles. An example of the issue can be seen in the attached screenshot.
![fa79546e-1b6a-11e7-966e-656cbec01aec](https://cloud.githubusercontent.com/assets/667850/24871316/0ef0d77c-1dce-11e7-80e2-0e6b29e99ff7.png)

sagarr:
I hope it also solves problem like this (road cuts at adjoining tiles)

![fireshot capture 1 - - http___localhost_8080_index html 16 39_49 58995_6 14185](https://user-images.githubusercontent.com/1051004/33139840-69570642-cfd4-11e7-9c53-e9826a80f648.png)

gdey:
Yes, I would say those issues are related to the title buffer. The buffer is not big enough.

gdey:
I no longer think the issue is because of the tile buffer is not big enough. I think this has to do with bounding boxes removing 
I no longer think it's because the tile buffer is not big enough. I think this has to do with the geometries we are grabbing before clipping. These geometries are the ones within a bounding box defined by the z/x/y tile. We then expand that bounding box the buffer size and clip the geometries to that new bounding box. Unfortunately, this means we don't grab all possible geometries that could be there in the newly expanded bounding box. (The Purple geometries in the figure.) I believe this is leading to the jagged edges we see as alluded to by [this StackOverflow post](https://stackoverflow.com/questions/40466547/vector-tiles-buffer). 
![explain_107](https://user-images.githubusercontent.com/14805/33532062-cba5ed4e-d84a-11e7-991f-01b4d6b02ada.png)

What we should be doing is expanding the bounding box by the buffer, and grab all the geometries in that new bounding box. We just clip to that bounding box as well.

gdey:
Anyone have an isolated dataset for this issue? I think I have a solution; that I need to test.

sagarr:
@gdey Do you have a solution on github branch which I can pull from? I want to test it.

gdey:
@sagarr Not yet, I'll push up as soon as I have figured out this bug, then ping you.

gdey:
@sagarr I just pushed up a branch called issue-107. Please try that branch out. I still have a few failing tests there, and some clean up to do. 

[...]

PR #225, opened 2017-12-14
Issue 107

gdey:
A major change to the way encoding and mvt generation works. Please look at #107 for details. A summary is that our understanding of buffers and want what gets drawn in them was slightly off. This fixes it so that all the geometries in a title (includes the buffer) is queried and drawn. This however required a bit of a rethink of how and more importantly where in the code these things were calculated and passed forward.


maths/makevalid/makevalid_test.go
```diff
@@ -10,8 +10,16 @@ import (
...
 	"github.com/terranodo/tegola/maths"
 	"github.com/terranodo/tegola/maths/hitmap"
+	"github.com/terranodo/tegola/maths/points"
 )
 
+const TileBuffer = 16
+
+var extent = points.Extent{
+	{0.0 - TileBuffer, 0.0 - TileBuffer},
+	{4096.0, 4096.0},
```
ARolek: should this have TileBuffer added as well?

maths/points/extent.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return true
+}
+
+// ContainsLine returns weather both points of the line are contained by the Extent.
+func (e Extent) ContainsLine(line [2][2]float64) bool {
+	return e.Contains(line[0]) && e.Contains(line[1])
+}
+
+// ContainsExtent returns waather the points of the second extent are containted by the first extent.
```
ARolek: minor type: waather -> weather

maths/points/extent.go
```diff
@@ -0,0 +1,110 @@
...
+func (e Extent) ContainsExtent(ee Extent) bool { return e.Contains(ee[1]) && e.Contains(ee[1]) }
+
+// Area returns the are of the Extent.
+func (e Extent) Area() float64 { return math.Abs(e[1][0]-e[0][0]) * (e[1][1] - e[0][1]) }
+
+// IntersectPt returns the intersect point if the one exists.
```
ARolek: minor typo: unnecessary "the"

mvt/feature.go
```diff
@@ -141,56 +138,33 @@ type cursor struct {
...
-
+// GetDeltaPointAndUpdate assumes the Point is in WebMercator.
 func (c *cursor) GetDeltaPointAndUpdate(p tegola.Point) (dx, dy int64) {
 	var ix, iy int64
-	if c.DisableScaling {
-		ix, iy = int64(p.X()), int64(p.Y())
-	} else {
-		ix, iy = c.ScalePoint(p)
+	var tx, ty = p.X(), p.Y()
+	// TODO: gdey — We should get rid of this, as we generally disable scaling; now.
```
ARolek: any reason to keep it? If not, lets purge it now.

mvt/feature_test.go
```diff
@@ -1,143 +1,128 @@
...
-				Miny: 0,
-				Maxx: 4096,
-				Maxy: 4096,
-			},
+	}
+	tile := tegola.NewTile(20, 0, 0)
+	fromPixel := func(x, y float64) *basic.Point {
+		pt, err := tile.FromPixel(tegola.WebMercator, [2]float64{x, y})
+		if err != nil {
+			panic(fmt.Sprintf("error trying to convert %v,%v to WebMercator. %v", x, y, err))
```
ARolek: why not t.Errorf(...)?

mvt/layer_test.go
```diff
@@ -129,31 +130,77 @@ func TestLayerAddFeatures(t *testing.T) {
...
+	tile := tegola.NewTile(0, 0, 0)
+
 	baseBBox := tegola.BoundingBox{
 		Minx: 0,
 		Miny: 0,
 		Maxx: 4096,
 		Maxy: 4096,
 	}
-	testcases := []struct {
+	fromPixel := func(x, y float64) *basic.Point {
```
ARolek: method is being repeated in TestEncodeGeometry. Maybe pull out of individual tests and put in mvt_test.go file?

provider/debug/debug.go
```diff
@@ -60,6 +60,39 @@ func (p *Provider) MVTLayer(ctx context.Context, layerName string, tile tegola.T
 			},
 		}
 		layer.AddFeatures(debugOutline)
+		ext1, err := tile.BufferedBoundingBox()
+		if err != nil {
+			return nil, err
+		}
+		/*
```
ARolek: purge this comment, or add a TODO to leverage the new log package.

provider/postgis/postgis_test.go
```diff
@@ -68,12 +68,8 @@ func TestMVTLayer(t *testing.T) {
...
 				},
 			},
-			tile: tegola.Tile{
-				Z: 1,
-				X: 1,
-				Y: 1,
-			},
-			expectedFeatureCount: 614,
+			tile:                 tegola.NewTile(1, 1, 1),
+			expectedFeatureCount: 4032,
```
ARolek: Did you test the new feature counts? Seems like a big jump by just adding the tile buffer.

tile.go
```diff
@@ -15,15 +24,75 @@ type Tile struct {
...
+	t.bufpext = [2][2]float64{{0 - t.Buffer, 0 - t.Buffer}, {t.Extent + t.Buffer, t.Extent + t.Buffer}}
+
 }
 
 func (t *Tile) Deg2Num() (x, y int) {
 	x = int(math.Floor((t.Long + 180.0) / 360.0 * (math.Exp2(float64(t.Z)))))
 	y = int(math.Floor((1.0 - math.Log(math.Tan(t.Lat*math.Pi/180.0)+1.0/math.Cos(t.Lat*math.Pi/180.0))/math.Pi) / 2.0 * (math.Exp2(float64(t.Z)))))
 
-	return
+	return x, y
```
ARolek: unnecessary since x,y are named returns.

tile.go
```diff
@@ -35,58 +104,171 @@ func (t *Tile) Num2Deg() (lat, lng float64) {
...
 		HasXYZ:  true,
 		X:       t.X,
 		Y:       t.Y,
 		Z:       t.Z,
-	}
+	}, nil
+}
+
+func (t *Tile) PixelBounds() (bounds [2][2]float64, err error) {
+	/*
```
ARolek: clean out this comment

tile.go
```diff
@@ -35,58 +104,171 @@ func (t *Tile) Num2Deg() (lat, lng float64) {
...
+		}
+		log.Println("Z", t.Z, "X", t.X, "Y", t.Y)
+		log.Println("Tile extent:", t.extent)
+		log.Println("Tile Pixel :", bounds)
+	*/
+	return [2][2]float64{{0.0, 0.0}, {t.Extent, t.Extent}}, nil
+}
+
+func (t *Tile) PixelBufferedBounds() (bounds [2][2]float64, err error) {
+	/*
```
ARolek: clean out this comment

tile_test.go
```diff
@@ -134,3 +115,32 @@ func TestTileZRes(t *testing.T) {
...
+			t.Errorf("[%v] Error, Expected nil Got %v", idx, err)
+			return
+		}
+		gpt, err := tile.ToPixel(tegola.WebMercator, npt)
+		if err != nil {
+			t.Errorf("[%v] Error, Expected nil Got %v", idx, err)
+			return
+		}
+		// TODO: gdey need to find the utility math function for comparing floats.
+		if pt[0] != gpt[0] {
```
ARolek: you can always use reflect.DeepEqual(), though it's a bit overkill for this use case.

atlas/atlas.go
```diff
@@ -47,7 +47,7 @@ func (a *Atlas) AllMaps() []Map {
 
 //	SeedMapTile will generate a tile and persist it to the
 //	configured cache backend
-func (a *Atlas) SeedMapTile(m Map, tile tegola.Tile) error {
+func (a *Atlas) SeedMapTile(m Map, tile *tegola.Tile) error {
```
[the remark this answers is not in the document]
gdey: Yes, we should I add nil checks.

basic/line.go
```diff
@@ -90,6 +90,14 @@ func NewLineFromSubPoints(points ...tegola.Point) (l Line) {
 	return l
 }
 
+func NewLineFrom2Float64(points ...[2]float64) (l Line) {
```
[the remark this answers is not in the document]
gdey: Yes, it's part of the translation to the geom types.

maths/makevalid/main.go
```diff
@@ -65,7 +63,33 @@ func destructure2(polygons [][]maths.Line, clipbox *points.BoundingBox) []maths.
 	return segments
 }
 
-func destructure5(ctx context.Context, hm hitmap.Interface, clipbox *points.BoundingBox, segments []maths.Line) ([][][]maths.Pt, error) {
+func MaxF64(vals ...float64) (max float64) {
```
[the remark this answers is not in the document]
gdey: This is a copy of the math.Max method made so that it is faster (by removing extra code.) I don't need to check to see if the value is NAN or Inf, as these are things I can assume from the construction of the function. Whereas math.Max is more generic and has to deal with overflows and special values.

maths/makevalid/makevalid_test.go
gdey: Yes, it should.

mvt/feature.go
gdey: It's still not sure how entangled it is. Once I've refactored the code for destructure5, I'll remove it. This is the reason for the TODO.

mvt/feature_test.go
gdey: Because this is an error with the test-case and not the function being tested. I want to panic, as the test case code is bad. And any result whether good or bad is suspect.

provider/debug/debug.go
gdey: The new log package will still compile this code, but will not remove it from the binary. It's really only needed when I'm debugging; this algo; It's really for the clipping issues I'm dealing with in issue-179. I'll remove it after I've proven that one out.

provider/postgis/postgis_test.go
gdey: Yes, this is from the test db.

tile.go
gdey: Yeah, but it still better to list the names. I not a fan of bare returns; as functions grow it becomes harder to know what is getting returned. I like naming returns because it gives documentation to the reader of the function what the function is returning. Especially when the type of the return is the same.

tile_test.go
gdey: This is to compare floats taking into account significant digits. DeepEqual would work just as well as comparing them directly. Right now this test could fail due to rounding errors; due to significant digits.

cmd/tegola/cmd/cache.go
```diff
@@ -239,10 +239,10 @@ var cacheCmd = &cobra.Command{
 		//	iterate our zoom range
 		for i := range zooms {
 
-			topLeft := tegola.Tile{Z: zooms[i], Long: bounds[0], Lat: bounds[1]}
+			topLeft := *tegola.NewTileLatLong(zooms[i], bounds[1], bounds[0])
```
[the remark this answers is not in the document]
gdey: No, it will never be nil.

cmd/tegola/cmd/cache.go
```diff
@@ -301,12 +301,7 @@ func parseTileString(str string) (tegola.Tile, error) {
 	if err != nil {
 		return tile, fmt.Errorf("invalid Y value (%v)", y)
 	}
-
-	tile = tegola.Tile{
-		Z: z,
-		X: x,
-		Y: y,
-	}
+	tile = tegola.NewTile(z, x, y)
```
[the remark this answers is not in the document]
gdey: I'll add a comment stating that to the NewTile functions.

mvt/layer_test.go
gdey: It is being used to help build out the test cases, instead of doing the math manually to convert from pixel coords to webmarcator for the lines, and polygons of a feature.

provider/debug/debug.go
gdey: NM will be removing this. Was thinking this different code. Will remove.

mvt/feature_test.go
gdey: Look at the code further, looks like I'm using it as a helper to build out the test cases, so returning an error is a no for the use case. Leaving for now, till I can think of a better system for building out the testcases.

cmd/tegola/cmd/cache.go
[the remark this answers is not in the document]
gdey: Sorry, I missed this change. I'll add it to issue-174

PR #234, opened 2017-12-21
max_connections value needs to be of type int64.

erictheise:
This throws an error as is.

```
2017/12/21 10:16:01 root.go:77: max_connections value needs to be of type int64. Value is of type string
```


ARolek:
Thank you!

PR #236, opened 2017-12-23
Issue 76 config bind and port

JivanAmara:
Rename current (string) 'port' config parameter to 'bind'.
Add int 'port' parameter (used in generating urls).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -37,7 +37,7 @@ type Config struct {
 type Webserver struct {
 	Bind              string `toml:"bind"`
 	HostName          string `toml:"hostname"`
-	Port              string `toml:"port"`
+	Port              int    `toml:"port"`
 	CORSAllowedOrigin string `toml:"cors_allowed_origin"`
 	LogFile           string `toml:"log_file"`
 	LogFormat         string `toml:"log_format"`
--- a/config/config.go
+++ b/config/config.go
@@ -35,9 +35,12 @@ type Config struct {
 }
 
 type Webserver struct {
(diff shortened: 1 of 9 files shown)


ARolek:

JivanAmara:
@ARolek Did this not make it into v0.6.0?

ARolek:
@JivanAmara I don't think it did. I will get the change into v0.7.0

config/config.go
```diff
@@ -35,9 +35,12 @@ type Config struct {
 }
 
 type Webserver struct {
+	Bind              string `toml:"bind"`
 	HostName          string `toml:"hostname"`
-	Port              string `toml:"port"`
+	Port              int    `toml:"port"`
```
ear7h: With Bind do we still need port and hostname?

ARolek:
After talking with @gdey we're going to change this behavior to:

- the `port` config will be converted from a string to a uint
- a new `bind` value will be added which will accept a string in the format of the current `port` config val.
- if both `port` and `bind` are set we throw an error during config validation.

gdey:
Looks like we found a way forward. But this never got accepted. Do we want to continue with this.

The way forward looks to be this:

`port` is going to be a uint
`bind` is a string.
both options are mutually exclusive

gdey:
Looking at the code, it may make sense to close this PR, open and issue and create a fresh patch. Code has change sigificantly since this PR was opened.

ARolek:
@gdey I think we should too. This is not a priority, so I think we close this PR for now and revist at a later date.

Issue #210, opened 2017-12-05
Support environment variables in config file


ARolek:
Several issues (#95, #133, #137) are requesting environment variable "injecting" into the config file. For example:

```
[[providers]]
name = "ne"
type = "postgis"
host = "{{.HOST}}"
port = {{.PORT}}
database = "{{.DATABASE}}"
user = "{{.USER}}"
password = "{{.PASSWORD}}"
```

We can accomplish this by using the go text/template package and reading in all the environment variables via os.Env().

ARolek:
An alternative approach is to implement the [toml.TextUnmarshller](https://godoc.org/github.com/BurntSushi/toml#TextUnmarshaler) interface for all our types to sniff for a prefix (i.e. the typical environment var $). This would avoid needing to use the Go template syntax and allow us to use a syntax that is more common place. For example:

```
[[providers]]
name = "ne"
type = "postgis"
host = "$HOST"
port = $PORT
database = "$DATABASE"
user = "$USER"
password = "$PASSWORD"
```

JivanAmara:
I like making the syntax match commonplace usage, why not simply make the replacements before the config goes to toml decoding?

ARolek:
@JivanAmara I'm already playing with a couple ideas, but I think the TextUnmarhshller interface is going to work well. We could read the config file prior to toml processing but I think we're going to shift the problem to regex processing, which is not necessarily a bad thing. Did you have another idea?

erictheise:
I mentioned leveraging the [Foreman](http://blog.daviddollar.org/2011/05/06/introducing-foreman.html) approach in https://github.com/terranodo/tegola-osm/issues/25. Three Go ports are listed in [Foreman's README.md](https://github.com/ddollar/foreman#ports).

Curious to hear @mojodna's latest thinking, too.

JivanAmara:
@ARolek I was thinking using regex processing to search & replace the environment vars in the config text (i.e. "port = $PORT" -> "port = 8080", then hand it off for decoding.  What do you mean by shifting the problem to regex processing?

JivanAmara:
@ARolek Something like this: https://github.com/terranodo/tegola/tree/issue-210_config_env_vars

ARolek:
@JivanAmara I dig the approach. I added a few comments, but overall I think this will work.

PR #237, opened 2017-12-24
Add support for env vars in configs. (#210)

JivanAmara:
Does a search and replace for $IDENTIFIER in config content and replaces with value from environment variables, or empty string if not set in environment.  After this sends the modified config content on for usual processing.


ARolek:
@JivanAmara I made a few changes to the branch. Please take a look and let me know any questions you have. Let me handle the merge please so I can get the release notes and change log updated.

config/config.go
```diff
@@ -138,25 +113,31 @@ func Parse(reader io.Reader, location string) (conf Config, err error) {
...
-		log.Printf("Problem reading from config reader: %v", err)
 		return nil, err
 	}
+
 	configStr := string(configBytes)
 
 	varPlaceHolders := varFinder.FindAllString(configStr, -1)
 	for _, ph := range varPlaceHolders {
 		// Get the environment variable value (drop the leading dollar sign ($))
 		envVal := os.Getenv(ph[1:])
```
gdey: Use `syscall.Getenv()` instead. It will tell you whether the variable exists or not.

ARolek: Good call. fixing coming up

config/config.go
```diff
@@ -136,6 +114,39 @@ func Parse(reader io.Reader, location string) (conf Config, err error) {
 	return conf, err
 }
 
+// replaceEnvVars replaces environment variable placeholders in reader stream with values
+// i.e. "val = $VAR" -> "val = 3"
+func replaceEnvVars(reader io.Reader) (io.Reader, error) {
+	// Variable definition follows IEEE Std 1003.1-2001
+	//   A dollar sign ($) followed by an upper-case letter, followed by
+	//   zero or more upper-case letters, digits, or underscores (_).
+	regexStr := `\$[A-Z]+[A-Z1-9_]*`
```
ARolek: @gdey brought up another approach, using [os.Expand](https://golang.org/pkg/os/#Expand) which would be cross platform and cleaner than the regex approach.

gdey: Yeah, looking at that now. I would say go with the regex, but enhance the regex to deal with `${SHELL_NAME}` as well. Or We require that all shell var substitution is done with `${}` to make it more obvious. In which case we would still want the regex.  (For reference: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html )

Issue #229, opened 2017-12-15
A GET request for a Tile with a negative row value is successful


jamesyarrington:
**To Reproduce:**
Send this GET Request:
```
<Tegola URL>/maps/osm/10/0/-1.pbf
```

**Expected Response:**
The server should return `400 Bad Request`.

**Actual Response:**
The Server returns `200 OK`.
This is the Tile object (converted using pbf, vector-tile):
```
VectorTile {
      layers:
       { 'debug-tile-outline':
          VectorTileLayer {
            version: 2,
            name: 'debug-tile-outline',
            extent: 4096,
            length: 1,
            _pbf: [Object],
            _keys: [Array],
            _values: [Array],
            _features: [Array] },
         'debug-tile-center':
          VectorTileLayer {
            version: 2,
            name: 'debug-tile-center',
            extent: 4096,
            length: 1,
            _pbf: [Object],
            _keys: [Array],
            _values: [Array],
            _features: [Array] } } }
```

_I do not know if the `debug-tile-outline` and `debug-tile-center` are from the vector-tile converter package, or from Tegola._

gdey:
The debug-tile-outline and debug-tile-center are added by Tegola. When you pass `?debug` to the inspector, a center point and the outline is drawn. These are the layers for those geometries.

PR #238, opened 2017-12-25
Updated server to check that x & y values for tiles are positive. (#229)

JivanAmara:
Resolves issue #229 & prevents similar issue with x value.


ARolek:
Looks like the map layer handler needed the checks as well. We should abstract that functionality to it's own method eventually so we don't have to keep these two pieces in sync.

PR #239, opened 2017-12-25
Issue 228 tile obtained outside proj.

### LINKED ISSUE #228: tegola.Tile returns a Tile even if given coordinates outside of the projection
**To Reproduce:**
Run this go program:
```go
package main

import (
	"fmt"

	tegola "github.com/terranodo/tegola"
)

func main() {
	zoom := 14
	var lon, lat float64
	lon = 0
	lat = 85.1
	tile := tegola.Tile{Z: zoom, Long: lon, Lat: lat}
	x, y := tile.Deg2Num()

	fmt.Printf("The x coodinate of this tile is %v.\n", x)
	fmt.Printf("The y coodinate of this tile is %v.\n", y)
}
```

**Expected Result:**
`tegola.Tile` should either throw an error, or return a Tile object within the bounds of the projection.

**Actual Result**
The program runs without error, returning a negative value for the y-coordinate.
```
The x coodinate of this tile is 8192.
The y coodinate of this tile is -26.
Success: process exited with code 0.
```

JivanAmara:
Resolves #228 
Updates Deg2Num() & Num2Deg to return (x, y, err) and check if values are within bounds for projection.  This prevents a tegola.Tile with invalid bounds from being propagated beyond these calls.

### DIFF
--- a/cmd/tegola/cmd/cache.go
+++ b/cmd/tegola/cmd/cache.go
@@ -240,10 +240,17 @@ var cacheCmd = &cobra.Command{
 		for i := range zooms {
 
 			topLeft := tegola.Tile{Z: zooms[i], Long: bounds[0], Lat: bounds[1]}
-			minx, maxy = topLeft.Deg2Num()
+			var errTL error
+			minx, maxy, errTL = topLeft.Deg2Num()
 
 			bottomRight := tegola.Tile{Z: zooms[i], Long: bounds[2], Lat: bounds[3]}
-			maxx, miny = bottomRight.Deg2Num()
+			var errBR error
+			maxx, miny, errBR = bottomRight.Deg2Num()
+
+			if errTL != nil || errBR != nil {
+				log.Printf("Problem(s) with Deg2Num() at zoom %v, skipping: %v / %v", zooms[i], errTL, errBR)
+				continue
+			}
 
 			//	range rows
 			for x := minx; x <= maxx; x++ {
--- a/cmd/tegola/cmd/cache.go
+++ b/cmd/tegola/cmd/cache.go
@@ -238,17 +238,20 @@ var cacheCmd = &cobra.Command{
 
 		//	iterate our zoom range
 		for i := range zooms {
+			var err error
 
 			topLeft := tegola.Tile{Z: zooms[i], Long: bounds[0], Lat: bounds[1]}
-			var errTL error
-			minx, maxy, errTL = topLeft.Deg2Num()
+			minx, maxy, err = topLeft.Deg2Num()
+			if err != nil {
+				log.Printf("error calculating Z/X/Y tile from zoom (%v), and lat / long (%v, %v) skipping: %v", zooms[i], topLeft.Lat, topLeft.Long, err)
+				continue
+			}
 
 			bottomRight := tegola.Tile{Z: zooms[i], Long: bounds[2], Lat: bounds[3]}
-			var errBR error
-			maxx, miny, errBR = bottomRight.Deg2Num()
 
-			if errTL != nil || errBR != nil {
-				log.Printf("Problem(s) with Deg2Num() at zoom %v, skipping: %v / %v", zooms[i], errTL, errBR)
+			maxx, miny, err = bottomRight.Deg2Num()
+			if err != nil {
+				log.Printf("error calculating Z/X/Y tile from zoom (%v), and lat / long (%v, %v) skipping: %v", zooms[i], bottomRight.Lat, bottomRight.Long, err)
 				continue
 			}
 
--- a/tile.go
+++ b/tile.go
@@ -1,6 +1,10 @@
 package tegola
 
-import "math"
+import (
+	"fmt"
+	"log"
+	"math"
+)
 
 const (
 	DefaultEpislon = 10.0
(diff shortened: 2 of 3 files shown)


tile.go
```diff
@@ -19,20 +23,39 @@ type Tile struct {
...
+	//	WebMercatorBounds = [4]float64{-20026376.39, -20048966.10, 20026376.39, 20048966.10}
 	n := math.Pi - 2.0*math.Pi*float64(t.Y)/math.Exp2(float64(t.Z))
 
 	lat = 180.0 / math.Pi * math.Atan(0.5*(math.Exp(n)-math.Exp(-n)))
 	lng = float64(t.X)/math.Exp2(float64(t.Z))*360.0 - 180.0
 
-	return lat, lng
+	if lat < -85.0511 || lat > 85.0511 || lng < -180.0 || lng > 180.0 {
+		msg := fmt.Sprintf("One or both outside valid range (x, y): (%v, %v)", t.X, t.Y)
+		err = fmt.Errorf("%v", msg)
```
gdey: No need to do `fmt.Sprintf(…)` and then `fmt.Errorf(…)`, combine the two:
`err = fmt.Errorf("One or both outside valid range (x, y): (%v, %v)", t.X, t.Y)`

Also, for errors, we are trying to standardize to starting with a lower letter for the message. i.e:
`err = fmt.Errorf("one or both are outside the valid range (x, y): (%v, %v)", t.X, t.Y)`

tile.go
```diff
@@ -19,20 +23,39 @@ type Tile struct {
...
 }
 
-func (t *Tile) Deg2Num() (x, y int) {
+func (t *Tile) Deg2Num() (x, y int, err error) {
+	// Converts from WGS84 Lat/Long postion to tile position (x,y)
+
+	// Check that input lat/long values are within WGS84 bounds
+	if t.Lat < -85.0511 || t.Lat > 85.0511 || t.Long < -180.0 || t.Long > 180.0 {
+		msg := fmt.Sprintf("One or both outside valid range (Long, Lat): (%v, %v)", t.Long, t.Lat)
+		err = fmt.Errorf("%v", msg)
```
gdey: No need to do fmt.Sprintf(…) and then fmt.Errorf(…), combine the two:
`err = fmt.Errorf("One or both outside valid range (Long, Lat): (%v, %v)", t.Long, t.Lat)`

Also, for errors, we are trying to standardize to starting with a lower letter for the message. i.e:
`err = fmt.Errorf("one or both are outside the valid range (Long, Lat): (%v, %v)", t.Long, t.Lat)`

cmd/tegola/cmd/cache.go
```diff
@@ -240,10 +240,17 @@ var cacheCmd = &cobra.Command{
...
+			var errTL error
+			minx, maxy, errTL = topLeft.Deg2Num()
 
 			bottomRight := tegola.Tile{Z: zooms[i], Long: bounds[2], Lat: bounds[3]}
-			maxx, miny = bottomRight.Deg2Num()
+			var errBR error
+			maxx, miny, errBR = bottomRight.Deg2Num()
+
+			if errTL != nil || errBR != nil {
+				log.Printf("Problem(s) with Deg2Num() at zoom %v, skipping: %v / %v", zooms[i], errTL, errBR)
```
gdey: Elsewhere we are `log.Fatal(err)` if there is an error. Should we not do the same here if there is a problem with the zoom level.

 @ARolek If the zoom is bad would it better to bail, and have the user fix the zoom level, instead of ignoring and skipping? I'm asking because I've not had to use the seed tool, and not sure what the automation around it is expecting.

gdey:
@ARolek When you get a chance, could you answer the above question about ignoring or using `log.Fatal`

tile.go
ARolek: In my experience the calculations are not perfect and therefore can produce values that are outside the range. I would prefer to not log.Fatal right now until we have a bullet proof calculation. Maybe if the value is out of range we could return the max value that's in range?

ARolek:
I'm working on merging this into v0.6.0 but it has effects that are propagating throughout the codebase based on a change that @gdey implemented to fix the tile buffer problem. We might need to reevaluate how `NewTile()` works or just deal with the fact that it can now throw an error.

tile.go
```diff
@@ -28,9 +43,10 @@ func (t *Tile) Deg2Num() (x, y int, err error) {
 
 	// Check that input lat/long values are within WGS84 bounds
 	if t.Lat < -85.0511 || t.Lat > 85.0511 || t.Long < -180.0 || t.Long > 180.0 {
-		err := fmt.Errorf("one or both outside valid range (Long, Lat): (%v, %v)", t.Long, t.Lat)
-		log.Print(err)
-		return -1, -1, err
+		return 0, 0, ErrInvalidLatLong{
```
JivanAmara: I suggest returning values that are obviously invalid.  There's no way to guarantee that the user of this will check the return value.  Values that are obviously invalid will help locate those cases.

tile.go
```diff
@@ -48,9 +64,10 @@ func (t *Tile) Num2Deg() (lat, lng float64, err error) {
 	lng = float64(t.X)/math.Exp2(float64(t.Z))*360.0 - 180.0
 
 	if lat < -85.0511 || lat > 85.0511 || lng < -180.0 || lng > 180.0 {
-		err := fmt.Errorf("one or both outside valid range (x, y): (%v, %v)", t.X, t.Y)
-		log.Print(err)
-		return -400.0, -400.0, err
+		return 0.0, 0.0, ErrInvalidXY{
```
JivanAmara: Same as above.

tile.go
ARolek: I think we should assume that the end user always handles errors. This is expected in Go (https://github.com/golang/go/wiki/CodeReviewComments#handle-errors), so handing back anything but the 0 value to me seems strange. I was going to ask you about those values and if they had significance. Thanks for chiming in on that change.

ARolek:
This change is having dramatic effects on the codebase now that @gdey has implemented a NewTile() method which calls these methods. There also appears to be issues with the calculations when X, Y or Z are 0 values which is very common. I think we should take a step back on this issue until we revisit how the calculations are being performed on the tiles if the NewTile() method currently in v0.6.0 is the correct approach. I have spent the last few hours working through this merge and it's getting worse and I worse. I fear that we're going to have major regressions in how tegola works unless we take a step back and think about a refactor on how tegola.Tile{} works.

JivanAmara:
I think a NewTile() constructor makes a lot of sense.  I think it would make a lot of sense to put the Num2Deg() & Deg2Num() checks in NewTile().  I don't understand why the changes to Num2Deg & Deg2Num cause such a problem, is it difficult for NewTile() to check their returns for errors?

ARolek:
The constructor is in v0.6.0. The problem is with any tiles that has a 0 in it for z,x or y. The Num2Deg() errors are kicking back that the value is invalid when it's not. I really think the tile struct is overloaded and those methods should be moved to a tilemath package where additional functions like Quadkey operators can also be implemented. A tile is suppose to be a slippy map tile which has Z/X/Y values. the tegola.TIle now has way too many concepts on it and the brittleness is starting to show. 

Without additional research into why 0 values for Z/X/Y are throwing errors on Num2Deg I can't merge this change and propagate it into the codebase.

ARolek:

ARolek:
Closing for now as this PR is stale.

Issue #231, opened 2017-12-15
Logger outputs file:line of log/standard.go along with timestamp in output.


JivanAmara:
This default output should be eliminated as the file & line number of the actual logging location are included in the output & this just confuses the output.

`2017/12/15 15:45:20 standard.go:23: •INFO•/home/jivan/.projects/tegola/src/github.com/terranodo/tegola/provider/gpkg/gpkgProvider.go•675•logging message`

PR #241, opened 2017-12-26
Update standard log backend to eliminate file:line from timestamp output. (#231)

JivanAmara:
Eliminates default prefix in standard log implementation so output matches expectation.


gdey:
LGTM

JivanAmara:
Added this commit to #222

PR #242, opened 2017-12-29
Issue 189 Tile BBox coordinate order

JivanAmara:
Adds tests for a number of conversion functions, verifying WebMercator values.
Swaps the position of Tile.BoundingBox() y-values so the smaller of the two is placed in Miny.

### DIFF
--- a/basic/g.go
+++ b/basic/g.go
@@ -15,7 +15,6 @@ func (g G) getbase() G {
 		bg, ok = g.Geometry.(G)
 	}
 	return rg
-
 }
 
 func (g G) IsLine() bool {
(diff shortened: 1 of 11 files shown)


ARolek:

JivanAmara:
Hey @ARolek, I vaguely remember you attending to this issue in v0.6.0.  Did that get completed, or do we still need to standardize bounding box coordinate order?

gdey:
I'm pretty sure that got resolved. I'm working on consolidating our various bbox and extent types. There was a bit of intermingling of two ideas bbox which made up of (min, max) points and what I need for some of my algos, (upper-left,lower-right), which needs to know about the coordinate system it's in and does not always translate to (min, max). So, that is being sorted. I think this issue can be closed.

ARolek:
@JivanAmara it's not fully resolved but did make some progress in v0.6.0. @gdey is working on plumbing this through but we're going to move to Bounds being `[4]float64{minx, miny, maxx, maxy}`

ARolek:
I think this PR should be closed and delete after PR #357 has been accepted.

Issue #245, opened 2017-12-29
Wire docker image build into CI.


JivanAmara:
After #244 gets completed, someone needs to update the build pipeline to include the docker images.

PR #247, opened 2018-01-03
Issue 244 - Dockerfile

JivanAmara:
Adds a Dockerfile for building a binary and creating a minimal deployment image using the alpine docker image as a base.

Using static linking, cgo support is included, entirely encapsulated in the binary with trivial difference to binary size.  Since it isn't affecting dependencies or binary/image size I think we should reconsider separating pure go & cgo-support docker images.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: go
+# Does nothing when building on terranodo/tegola, but makes builds work cleanly on forks.
+go_import_path: github.com/terranodo/tegola
 
 go:
   - 1.9.x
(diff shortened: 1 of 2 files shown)


JivanAmara:
Changes for issue #245 are small and closely related, so included here.

@gdey, @ARolek, I currently have Travis configured to push the docker image to jivanamara/tegola.  I'll need either:
1. Permission for dockerhub user jivanamara to write to terranodo/tegola or
1. The travis-secured DOCKER_USER & DOCKER_PASSWORD for a generic tegola user with those permissions.

Once I've got one of those, it is a couple minor changes to the Travis config in order to push a terranodo/tegola:<tag> image to DockerHub on releases.

ingenieroariel:
Jivan,

I have added you as an owner to the Terranodo org on Dockerhub.

One question about `/opt/tegola_config`. What do you think about just putting things on the root so that you have `/tegola`, `/config.toml` and `/data.gpkg`? There is nothing else running on that container (and nothing else will run in the future) so we can safely ignore any linux distro conventions.

JivanAmara:
Hey Ariel, I'm not attached to where the binary goes in the end, but the
config & gpkg have to be in a different directory since they're supplied
via a volume.

JivanAmara:
Alright, this has been tested in the wild on branch master & is working properly.  Ready to merge.

PR #248, opened 2018-01-04
Issue 107

ARolek:
This PR finalizes the requirements for issue-107: configurable tile buffer.


tile.go
```diff
@@ -8,10 +8,10 @@ import (
 )
 
 const (
-	DefaultEpislon    = 10.0
-	DefaultExtent     = 4096
-	DefaultTileBuffer = 64
-	MaxZ              = 20
+	DefaultEpislon            = 10.0
+	DefaultExtent             = 4096
+	DefaultTileBuffer float64 = 64
```
gdey: What's the reason for adding the float64 here?
`Consts` generally don't have a type associated with them.

ARolek: I did it to be explicit since tile.Buffer expects a float64 but this const will be inferred as an int64 unless we change it to 64.0, which I'm also fine with.

gdey: Let's change it to 64.0. I think that is better.

ARolek: done

Issue #288, opened 2018-02-04
Add GeoJSON encoding to geom package.


JivanAmara:
Primary interest is for returning GeoJSON from wfs service, but will probably be useful elsewhere.

PR #289, opened 2018-02-04
Add GeoJSON encoding to geom package. (#288)

JivanAmara:
Title says it all.


geom/encoding/geojson/geojson.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type Geometry struct {
+	geom.Geometry
+}
+
+func (geo Geometry) MarshalJSON() ([]byte, error) {
+	type coordinates struct {
+		Type  GeoJSONType `json:"type"`
+		Coord interface{} `json:"coordinates,omitempty"`
```
JivanAmara: `Coords` to be consistent with `collection.Geometries`

geom/encoding/geojson/geojson.go
```diff
@@ -0,0 +1,122 @@
...
+		return json.Marshal(collection{
+			Type:       GeometryCollectionType,
+			Geometries: geos,
+		})
+	}
+
+	return nil, ErrUnknownGeometry{geom.Geometry(geo)}
+}
+
+type featureType struct{}
```
JivanAmara: Add comment briefly explaining the need for this type.

geom/encoding/geojson/geojson.go
```diff
@@ -0,0 +1,122 @@
...
+type Feature struct {
+	Type featureType `json:"type"`
+	ID   *uint64     `json:"id,omitempty"`
+	// can be null
+	Geometry Geometry `json:"geometry"`
+	// can be null
+	Properties map[string]interface{} `json:"properties"`
+}
+
+type featureCollectionType struct{}
```
JivanAmara: Add comment briefly explaining the need for this type.

geom/encoding/geojson/geojson_test.go
```diff
@@ -0,0 +1,115 @@
...
+		t.Parallel()
+
+		f := geojson.Feature{
+			Geometry: geojson.Geometry{tc.geom},
+		}
+
+		output, err := json.Marshal(f)
+		if err != nil {
+			t.Error(err)
+			return
```
JivanAmara: Let the test continue to see in one shot if other test cases fail?

gdey: It seems all these tests are expecting good cases. In that case, I think it makes more sense to bail if there is an error Marshalling. 

It would be better to add an error element to the `tcase struct`, so that we can have bad cases.

geom/encoding/geojson/geojson.go
gdey: Just to clarify change `Coord` to `Coordinates`. If so 👍

geom/encoding/geojson/geojson_test.go
ARolek: I will add an invalid geom test too.

ARolek:
updates are in.

JivanAmara:
Looks good to me, can't approve since I opened the PR.

Issue #291, opened 2018-02-07
geom/encoding/wkt does not handle nil and empty geometries correctly.


gdey:
The new implementation of wkt does not handle nil values and empty geometries.
Add test cases as necessary.
Fix any issues.

PR #294, opened 2018-02-08
Issue 291 wkt nil

gdey:
Fixed issue #291 where wkt encoder will panic for nil or zero value geometries.


ARolek:
@gdey this commit looks good to me. Can you just add a Readme to the fuzz package with some instructions on how to use it?

PR #308, opened 2018-02-18
Gpkg autoconfig

JivanAmara:
Adds an AutoConfig() function to gpkg provider.  The config includes all available feature tables.  This function provides what NewTileProvider() expects, not a config.toml.

Also extracted out metadata collection from NewTileProvider() into featureTableMetaData() so it can be reused in AutoConfig() & more rigorously tested.


JivanAmara:
No changes, just rebased onto current v0.6.0.

ARolek:

JivanAmara:
Rebased onto v0.7.0

ARolek:
I'm fine with this PR for it's near term purpose. AutoConfig is a great feature that I would like to eventually establish an interface around, but that should be handled in another issue.

Issue #300, opened 2018-02-13
Add support for a redis caché, to the caché system.


gdey:
Moving over feature request from tegola-osm/#40 for @peldhose .

peldhose:
ok thanks ... 👍
how much time it may take to implement feature ?? any roadmap or timeline??

ARolek:
@peldhose we have a few other priorities so it's not on the immediate roadmap. Would you be interested in trying to build the implementation? I can point you to where in the code you would do the work.

peldhose:
mmm ok , actually i coded years before in java n php.anyway give me some guidelines i will try myself to implement (u guys should help me ).
also we would like to use tegola for production purpose, any suggestions ??

gdey:
@peldhose If you have any questions ask, and I'll do my best to help you out. To learn Go you can start with: 
https://tour.golang.org/welcome/1

As far as implementing the caché:
Here is are a couple popular redis libs: 
https://godoc.org/github.com/garyburd/redigo/redis
https://godoc.org/github.com/go-redis/redis

I'm not sure which is better. 

To add a new caché, it will need to register with the caché system located here:
https://github.com/terranodo/tegola/tree/master/cache

Looking at the file caché you can see it doing the registration with the caché system here:
https://github.com/terranodo/tegola/blob/master/cache/filecache/filecache.go#L24,26

Each caché must the caché interface:
https://github.com/terranodo/tegola/blob/master/cache/cache.go#L12-16

So, all you have to do it implement the caché interface for the storing, retrieving, and purging. Register with the registration function, which you provide a name and an InitFunc (https://github.com/terranodo/tegola/blob/master/cache/cache.go#L112) that will take a configuration `map[string]interface{}` and return the caché object.  The name is what user use in the configuration file, so it is good to turn that into a constant (https://github.com/terranodo/tegola/blob/master/cache/filecache/filecache.go#L17).  

You can use the filecache ( https://github.com/terranodo/tegola/tree/master/cache/filecache ) as a guide. 

Also, base your work on the (v0.6.0 branch). The code is better commented as well, and it will soon be master.

If you have any further questions don't hesitate to ask.

peldhose:
ok, let me go through n need some time to understand GO way ... anyway, i will try my best.

 we would like to use tegola for production purpose, any suggestions or concerns i should consider  ??

ear7h:
@peldhose if you haven't gone through too much trouble already I could implement this

peldhose:
[...]

PR #310, opened 2018-02-19
Issue 300 Redis caché

ear7h:
As requested in issue #300

The maxZoom mechanism I implemented is different than the other caches. I think the other caches are only caching the higher zoom levels. I will open a separate issue for this.


ARolek:
@ear7h nice! Thanks for tacking this guy. The travis check is failing initially because the package you rely on has not been vendored. Can you use `govendor` and then push an update? I will review the driver's license in the mean time.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	for i, tc := range testcases {
+		_, err := rediscache.New(tc.config)
+		if err != nil {
+			if tc.err != nil && err.Error() == tc.err.Error() {
+				//	correct error returned
+				continue
+			}
+			t.Errorf("testcase (%v) failed. err: %v", i, err)
```
gdey: We have been moving our test error reporting to a standardized format. 

The format takes the form of:

` '[' $testcase_number '] ' $test_subject ', expected ' $expected_value ' got ' $gotten_value` 

So, for the above it would look like:
```
t.Errorf("[%v] unexpected err, expected %v got %v", i, tc.err, err)
```

However, moving to the subtest approach is preferred. It's a bit more of a change, you can find an example here: https://github.com/terranodo/tegola/blob/v0.6.0/geom/cmp/cmp_test.go

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			},
+			expectedData: nil,
+			expectedHit:  true,
+		},
+	}
+
+	for i, tc := range testcases {
+		rc, err := rediscache.New(tc.config)
+		if err != nil {
+			t.Errorf("testcase (%v) failed. err: %v", i, err)
```
gdey: See note about the error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	test write
+		if tc.expectedHit {
+			err = rc.Set(&tc.key, tc.expectedData)
+			if err != nil {
+				t.Errorf("testcase (%v) write failed. err: %v", i, err)
```
gdey: See Note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+			continue
+		}
+
+		// test read
+		output, hit, err := rc.Get(&tc.key)
+		if err != nil {
+			t.Errorf("testcase (%v) read failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+		if tc.expectedHit == hit {
+			t.Errorf("testcase (%v) read failed. should hit should have been %b but cache reported a %b", i, tc.expectedHit, hit)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		if !reflect.DeepEqual(output, tc.expectedData) {
+			t.Errorf("testcase (%v) failed. output (%v) does not match expected (%v)", i, output, tc.expectedData)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	test purge
+		if tc.expectedHit {
+			err = rc.Purge(&tc.key)
+			if err != nil {
+				t.Errorf("testcase (%v) failed. purge failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+				continue
+			}
+		} else {
+			// test purge non-existent key
+			err = rc.Purge(&tc.key)
+			if err != nil {
+				t.Errorf("testcase (%v) failed. purge failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			bytes1:   []byte("\x66\x6f\x6f"),
+			bytes2:   []byte("\x53\x69\x6c\x61\x73"),
+			expected: []byte("\x53\x69\x6c\x61\x73"),
+		},
+	}
+
+	for i, tc := range testcases {
+		rc, err := rediscache.New(tc.config)
+		if err != nil {
+			t.Errorf("testcase (%v) failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	test write1
+		if err = rc.Set(&tc.key, tc.bytes1); err != nil {
+			t.Errorf("testcase (%v) write failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	test write2
+		if err = rc.Set(&tc.key, tc.bytes2); err != nil {
+			t.Errorf("testcase (%v) write failed. err: %v", i, err)
```
gdey: See note about error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	fetch the cache entry
+		output, hit, err := rc.Get(&tc.key)
+		if err != nil {
+			t.Errorf("testcase (%v) read failed. err: %v", i, err)
```
gdey: See note above about the error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+		if !hit {
+			t.Errorf("testcase (%v) read failed. should have been a hit but cache reported a miss", i)
```
gdey: See note above about the error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		if !reflect.DeepEqual(output, tc.expected) {
+			t.Errorf("testcase (%v) failed. output (%v) does not match expected (%v)", i, output, tc.expected)
```
gdey: See note above about the error message format.

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		//	clean up
+		if err = rc.Purge(&tc.key); err != nil {
+			t.Errorf("testcase (%v) failed. purge failed. err: %v", i, err)
```
gdey: See note above about the error message format.

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	defaultAddress := "127.0.0.1:6379"
+	defaultPassword := ""
+	defaultDB := 0
+	defaultMaxZoom := 0
+
+	c := dict.M(config)
+
+	network, err := c.String(ConfigKeyNetwork, &defaultNetwork)
+	if err != nil {
+		return
```
gdey: Since this a pretty long function, please don't use a bare return. 
Especially since it is easy to do a 
`return nil, nil` 
here.

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	addr, err := c.String(ConfigKeyAddress, &defaultAddress)
+	if err != nil {
+		return
```
gdey: Please, don't use the bare return here. 
`return nil, err`

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	password, err := c.String(ConfigKeyPassword, &defaultPassword)
+	if err != nil {
+		return
```
gdey: change to the non-bare version:
`return nil, err`

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	db, err := c.Int(ConfigKeyDB, &defaultDB)
+	if err != nil {
+		return
```
gdey: change to non-bare version:
`return nil, err`

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+		Addr:        addr,
+		Password:    password,
+		DB:          db,
+		PoolSize:    2,
+		DialTimeout: 3 * time.Second,
+	})
+
+	pong, err := client.Ping().Result()
+	if err != nil || pong != "PONG" {
+		return
```
gdey: change to the non-bare version:
`return nil, err`

Should this return an error if pong != "PONG". What does that mean? We are sending back a nil for the client, but also, a nil error here? How, is the client suppose to deal with this?

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return
+	}
+
+	rcache = &RedisCache{
+		Redis:   client,
+		MaxZoom: maxZoom,
+	}
+
+	return
```
gdey: Don't use a bare return here.
Please, change this to:
```
return &RedisCache{
        Redis: client,
        MaxZoom: maxZoom,
}, nil
```

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+
+	pong, err := client.Ping().Result()
+	if err != nil || pong != "PONG" {
+		return
+	}
+
+	maxZoom, err := c.Int(ConfigKeyMaxZoom, &defaultMaxZoom)
+	if err != nil {
+		return
```
gdey: Please, change the bare return to:
`return nil, nil`

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+	val, err = rdc.Redis.Get(key.String()).Bytes()
+
+	if err == nil { // cache hit
+		hit = true
+	} else if err == redis.Nil { // cache miss
+		err = nil // clear error
+	}
+
+	return
+}
```
gdey: This is hard for me to read:
```
func (rdc *RedisCache) Get(key *cache.Key) (val []byte, hit bool, err error){
           val, err = rdc.Redis.Get(key.String()).Bytes()
           switch err {
           case nil: // caché hit
                     return val, true, nil
           case redis.Nil: // caché miss
                     return val, false, nil
           default: // error 
                     return nil, false, err
          }
}
```

I feel is easier to read, as it immediately obvious what is happening.

cache/rediscache/rediscache.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (rdc *RedisCache) Purge(key *cache.Key) (err error) {
+	err = rdc.Redis.Del(key.String()).Err()
+	return
+}
```
gdey: I would change this to just:
```
func (rdc *RedisCache) Purge(key *cache.Key) error {
        return rdc.Redis.Del(key.String()).Err()
}
```

cache/rediscache/rediscache.go
ear7h: This should be changed. The redis `ping` command should return the string `PONG`, receiving something else shouldn't happen.

ear7h:
@ARolek I was meaning to ask how to handle the dependency, I'll include it.

@gdey I'll work on the styling and use explicit returns

cache/rediscache/rediscache.go
```diff
@@ -63,21 +64,22 @@ func New(config map[string]interface{}) (rcache cache.Interface, err error) {
 	})
 
 	pong, err := client.Ping().Result()
-	if err != nil || pong != "PONG" {
-		return
+	if err != nil {
+		return nil, err
+	}
+	if pong != "PONG" {
+		return nil, fmt.Errorf("redis did not resoind with 'PONG', '%s'", pong)
```
gdey: Typo: `resoind` should it be `respond`?

ARolek:
looks like you vendored the redis client but forgot to add it to your commit. i see the changes to the vendor.json file but the package is not included. 

Also let's update the travis config to [include redis](https://docs.travis-ci.com/user/database-setup/#Redis) and make sure the env var is set to run the tests. Thanks again for tackling this!

cache/rediscache/rediscache.go
ear7h: That is correct, I'll change this

ear7h:
@ARolek ah okay, I've never used govendor. I thought the vendor directory was treated like `node_modules` I'll add it

cache/rediscache/redicache_test.go
```diff
@@ -0,0 +1,204 @@
...
+			continue
+		}
+
+		// test read
+		output, hit, err := rc.Get(&tc.key)
+		if err != nil {
+			t.Errorf("[%v] read failed with error, expected %v got %v", k, nil, err)
+			continue
+		}
+		if tc.expectedHit == hit {
```
gdey: Should this be `tc.expectedHit != hit` ?

ear7h: Yes it should

ear7h:
The failed test [here](https://travis-ci.org/terranodo/tegola/jobs/343959092#L710) is a result of different go versions returning different error messages for the same error. Should I use a `switch` on the go version? Or is there another way to handle this.

ARolek:
@ear7h I think the error is type [OpError](https://golang.org/pkg/net/#OpError). You could create an OpError instance and fill it with the values you expect then compare the Error() string outputs of both errors. This should work across Go versions.

Additionally, now that Go 1.10 is out, we're going to drop support for Go 1.7.x as we target the 3 newest versions.

cache/rediscache/rediscache_test.go
```diff
@@ -0,0 +1,204 @@
...
+	if os.Getenv("RUN_REDIS_TESTS") != "yes" {
+		fmt.Println("RUN_REDIS_TESTS not set to 'yes' skipping tests")
+		return
+	}
+
+	type tc struct {
+		config map[string]interface{}
+		err    error
+	}
+
```
gdey: For this issue: 
> The failed test [here](https://travis-ci.org/terranodo/tegola/jobs/343959092#L710) is a result of different go versions returning different error messages for the same error.  Should I use a switch on the go version? Or is there another way to handle this.

I would change the test case to the following:
```
type tc struct {
     config map[string]interface{}
     errMatch string
}
```

cache/rediscache/rediscache_test.go
```diff
@@ -0,0 +1,204 @@
...
+	testcases := map[string]tc{
+		"redis explicit config": {
+			config: map[string]interface{}{
+				"network":  "tcp",
+				"address":  "127.0.0.1:6379",
+				"password": "",
+				"db":       0,
+				"max_zoom": 0,
+			},
+			err: nil,
```
gdey: Then here for `err: nil` I would replace with
` errMatch: ""` or better leave it out.

cache/rediscache/rediscache_test.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+		"redis implicit config": {
+			config: map[string]interface{}{},
+			err:    nil,
```
gdey: Then here for `err: nil` I would replace with
` errMatch: ""` or better leave it out.

cache/rediscache/rediscache_test.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+		"redis bad config":{
+			config: map[string]interface{}{
+				"address": "127.0.0.1:6000",
+			},
+			err: fmt.Errorf("dial tcp 127.0.0.1:6000: connect: connection refused"),
```
gdey: Here `err: fmt.Errorf("…"),` would change to:
`errMatch: "connection refused",`

cache/rediscache/rediscache_test.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+	}
+
+	for i, tc := range testcases {
+		_, err := rediscache.New(tc.config)
+		if err != nil {
+			if tc.err != nil && err.Error() == tc.err.Error() {
```
gdey: this would change to:
` if tc.errMatch != "" && strings.Contains(err.Error(),tc.errMatch) {`

doc for [strings.Contains](https://godoc.org/strings#Contains)

ear7h: Ah okay, that should fix it. Thank you.

gdey:
@ear7h @ARolek just informed me that we are still missing the MaxZoom changes. Could you get those in? I'm not sure what exactly the change needs to be.

ear7h:
@gdey Yea I'll do that. I was also gonna refactor the `cache.Key` struct to use `uint` for X/Y/Z values (#312 ), should it be included in this pull request?

cc @ARolek

ARolek:
@ear7h I would say send in #312 in a different PR. I'm working towards releasing v0.6.0 this week and it would be cool to have the redis cache included. I don't want it to be blocked by the uint conversion since that's a clean up task.

Issue #311, opened 2018-02-19
Caching at higher levels than specified by maxZoom 


ear7h:
From what I understand [here](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels) higher zooms correspond with higher numbers. However the code in the `filecache` and `s3cache` packages assume the opposite, not caching when the zoom level is lower than `MaxZoom`.

```go
// inside Set() method

//	check for maxzoom
if fc.MaxZoom != nil && key.Z <= int(*fc.MaxZoom) {
	return nil
}

// do cache
```

Also, I think mechanism is makes odd use of pointers and uint types. I'm willing to change this but wanted to submit an issue for clarification.

ARolek:
@ear7h you're correct about "lower zooms" being closer to zero, and this looks like a bug. Thanks for reporting it. I will add a test case to both the `filecache` and `s3cache` and get this fixed.

Regarding the uint / pointer piece being odd, I agree. This is because some of the typing is going through a transition. Tile requests come to tegola in the format of `/:map_name/:z/:x/:y`. Currently we parse the z/x/y values into `int` values, but then we hit a bug with invalid requests of negative values. You can see the checking happening [here](https://github.com/terranodo/tegola/blob/master/server/handle_map_zxy.go#L51). The plan is to convert everything to `uint` values and drop the less than zero checking. This will also require an update to the cache interface and then some code cleanup to the various cache implementations. I will open another issue to track it.

ear7h:
Okay sounds good. I'll look for those tests when you submit them and make corresponding ones for the `rediscache`

ARolek:
@ear7h my recent PR (https://github.com/terranodo/tegola/pull/314) has the max_zoom fix and associated tests. Thanks again for the report. Teach me to not write test cases... ;-)

PR #314, opened 2018-02-21
max_zoom bug on s3 and file caches #311

ARolek:
- added test cases for cache max_zoom
- removed package stutter cache/s3cache -> cache/s3
- removed package stutter cache/filecache -> cache/file
- updated test cases to use new convetion


.gitignore
```diff
@@ -1,4 +1,4 @@
 .DS_Store
 .idea
 documents
-*.toml
```
gdey: What's the reason for the removal of the new line?

ARolek: I was editing the .gitignore but decided against the entries. This was purely an artifact of that modification and subsequent manual reversion.

Issue #320, opened 2018-02-23
error fetching geometry type for layer


pnorman:
With this config I get an error

```toml
[webserver]

[[providers]]
name = "osm2pgsql"
type = "postgis"
host = "/var/run/postgresql" # Default for Debian-based distributions
port = 5432                     # postgis database port
database = "gis"        # postgis database name
user = "pnorman"
password = ""

  [[providers.layers]]
  name = "water"
  id_fieldname = "osm_id"
  geometry_fieldname = "way"
  sql = """
  SELECT
      0::bigint AS osm_id,
      ST_AsBinary(way) AS way
    FROM ne_ocean
    WHERE way && !BBOX!
  UNION ALL
  SELECT
      0::bigint AS osm_id,
      ST_AsBinary(way) AS way
    FROM ne_lake
    WHERE way && !BBOX!
  """

[[maps]]
name = "bolder"

[[maps.layers]]
provider_layer = "osm2pgsql.water"
min_zoom = 0
max_zoom = 5
```

```
$ SQL_DEBUG=EXECUTE_SQL ./tegola_linux_amd64 serve
2018/02/22 17:02:44 config.go:161: Loading local config (config.toml)
2018/02/22 17:02:44 root.go:75: error fetching geometry type for layer (water): ERROR: UNION types text and bytea cannot be matched (SQLSTATE 42804)
```

pnorman:
My error - wrong function, and my eyes saw what I was expecting, not what I had typed.

pnorman:
So the SQL I wrote has two calls of ST_AsBinary(way)

tegola issued this SQL to the server:

```sql
select
      0::bigint as osm_id,
      st_geometrytype(way) as way
    from ne_ocean
    where way && st_makeenvelope(-2.003750834e+07,2.003750834e+07,2.003750834e+07,-2.003750834e+07,3857)
  union all
  select
      0::bigint as osm_id,
      st_asbinary(way) as way
    from ne_lake
    where way && st_makeenvelope(-2.003750834e+07,2.003750834e+07,2.003750834e+07,-2.003750834e+07,3857)
   LIMIT 1
```

Is Tegola trying to parse the SQL as text and modify it to issue a query?

gdey:
Yes, it is trying to use regular expressions to do the `!token!` substitution, and for some of the feature discovery, it tries to modify the SQL. It's rudimentary right now, and I have plans to implement a more robust SQL parser to enable us to be more intelligent about these transforms.

[...]

PR #324, opened 2018-02-23
Replace all occurances of ST_AsBinary when checking types

pnorman:
There are issues with checking types using ST_GeometryType that this doesn't fix, but this does fix the issue of UNIONed tables and mixed types.

In addition to this applying to 0.5.x, it should be backported to 0.4.x, where the regression was introduced.

It fixes the **specific** error in #320 but that one should be left open or a new issue created about the problems of relying on string replacement like this.

### DIFF
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -272,7 +272,7 @@ func (p Provider) layerGeomType(l *Layer) error {
 
 	//	we want to know the geom type instead of returning the geom data so we modify the SQL
 	//	TODO: this strategy wont work if remove the requirement of wrapping ST_AsBinary(geom) in the SQL statements.
-	sql = strings.Replace(strings.ToLower(sql), "st_asbinary", "st_geometrytype", 1)
+	sql = strings.Replace(strings.ToLower(sql), "st_asbinary", "st_geometrytype", -1)
 
 	//	we only need a single result set to sniff out the geometry type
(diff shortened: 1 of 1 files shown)


gdey:
@pnorman could you base this pull request off of v0.6.0. v0.6.0 is the head of our development; not master. For us master is the current stable.

ARolek:
@pnorman good catch! I'm going to spin this into another issue and base it off v0.6.0. I still need to respond to your issue around SQL parsing. One detail you brought up about not using strings.ToLower() has been fixed in the v0.6.0 branch which we're hoping to cut today. Thanks for submitting this fix

ARolek:
@pnorman I just reviewed the code in v0.6.0 and this has been fixed. You can see the fixes [here](https://github.com/terranodo/tegola/blob/v0.6.0/provider/postgis/postgis.go#L271).

pnorman:
> @pnorman could you base this pull request off of v0.6.0. v0.6.0 is the head of our development; not master. For us master is the current stable.

The intent for this is for 0.5.1, not 0.6.0

ARolek:
@pnorman we're so close to cutting v0.6.0 and this has been fixed in v0.6.0, so let's hold off on the patch release.

pnorman:
Even if 0.6.0 were out, as a bug fix shouldn't it make it into 0.5.x?

Issue #327, opened 2018-02-23
Add closing coordinate to GeoJSON Polygons.


JivanAmara:
GeoJSON requires that for closed LineString types the final coordinate is the same as the first.  Our (and others', incl WKBGeometry/MapBox Vector Tile) geometries' representations don't require this. The geojson encoder needs to be updated to add this final coordinate for linestrings in Polygon/MultiPolygon types.

JivanAmara:
^ @ARolek

PR #329, opened 2018-02-23
close polygon rings for geojson #327

ARolek:
- close polygon (and multi polygon) rings during MarshalJSON calls
to match GeoJSON spec.

### DIFF
--- a/geom/encoding/geojson/geojson.go
+++ b/geom/encoding/geojson/geojson.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/terranodo/tegola/geom"
 	"github.com/terranodo/tegola/geom/encoding"
+	"github.com/terranodo/tegola/internal/log"
 )
 
 type GeoJSONType string
@@ -132,18 +133,22 @@ type FeatureCollection struct {
 // is not the same it will add an extra point to "close" the polygon per the GeoJSON spec.
 func closePolygon(p geom.Polygoner) geom.Polygoner {
 	rings := p.LinearRings()
-	for i := range rings {
 
+	rs := make([][][2]float64, 0, len(rings))
+	for i := range rings {
 		if len(rings[i]) < 3 {
-			return geom.Polygon(rings)
+			log.Warn("encounted polygon with less than 3 points. dropping")
+			continue
 		}
 
 		//	check if the first point and the last point are the same
 		//	if they're not, make a copy of the first point and add it as the last position
 		if rings[i][0] != rings[i][len(rings[i])-1] {
-			rings[i] = append(rings[i], rings[i][0])
+			rs = append(rs, append(rings[i], rings[i][0]))
+		} else {
+			rs = append(rs, rings[i])
 		}
 	}
 
-	return geom.Polygon(rings)
+	return geom.Polygon(rs)
 }
--- a/geom/encoding/geojson/geojson.go
+++ b/geom/encoding/geojson/geojson.go
@@ -60,25 +60,26 @@ func (geo Geometry) MarshalJSON() ([]byte, error) {
 		})
 
 	case geom.Polygoner:
+		ps := g.LinearRings()
+		closePolygon(ps)
+
 		return json.Marshal(coordinates{
 			Type: PolygonType,
 			//	make sure our rings are closed
-			Coords: closePolygon(g).LinearRings(),
+			Coords: ps,
 		})
 
 	case geom.MultiPolygoner:
 		ps := g.Polygons()
 
-		//	pre allocate our memory for the copy
-		var mp = make(geom.MultiPolygon, 0, len(ps))
 		//	iterate through the polygons making sure they're closed
-		for _, p := range ps {
-			mp = append(mp, closePolygon(geom.Polygon(p)).LinearRings())
+		for i := range ps {
+			closePolygon(geom.Polygon(ps[i]))
 		}
 
 		return json.Marshal(coordinates{
 			Type:   MultiPolygonType,
-			Coords: mp.Polygons(),
+			Coords: ps,
 		})
 
 	case geom.Collectioner:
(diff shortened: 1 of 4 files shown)


geom/encoding/geojson/geojson.go
```diff
@@ -117,3 +127,23 @@ type FeatureCollection struct {
...
+
+		//	check if the first point and the last point are the same
+		//	if they're not, make a copy of the first point and add it as the last position
+		if rings[i][0] != rings[i][len(rings[i])-1] {
+			rings[i] = append(rings[i], rings[i][0])
+		}
+	}
+
+	return geom.Polygon(rings)
+}
```
gdey: We don't want to abandon all the rings after rings with less then 3 points. I think you should return a new slice.
```
fun closePolygon(p geom.Polygoner) geom.Polygoner {
      rings := p.LinearRings()
       r = make([][][2]float64, 0, len(rings))
      for i := range rings {
                   if len(rings[i]) < 3 { 
                           continue
                  }
                  // check if the first point and the last point are the same, if not make them.
                  if rings[i][0] != rings[i][len(rings[i])-1] {
                       r = append(r, append(rings[i],rings[i][0]))
                  } else {
                      r = append(r, rings[i])
                 }
            }
           return geom.Polygon(r)
}
```

The way this is done should have the same memory foot print as what you had before.

geom/encoding/geojson/geojson_internal_test.go
```diff
@@ -0,0 +1,58 @@
...
+				},
+			},
+			expected: geom.Polygon{
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3.2, 4.3},
+				},
+			},
+		},
+	}
+
```
gdey: Add the following test cases:
```go
"two point only polygon": {
 			geom: geom.Polygon{
 				{
 					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
 				},
 			},
 			expected: geom.Polygon{},
 		},
"two point  polygon 1": {
 			geom: geom.Polygon{
                               { geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3, 4} },
 				{	geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5} },
 			},
 			expected: geom.Polygon{
 { geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3, 4} },
},
 		},

"two point only polygon 2": geom: geom.Polygon{
{	geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5} },
                               { geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3, 4} },
 			},
 			expected: geom.Polygon{
 { geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3, 4} },
},
 		},
```

geom/encoding/geojson/geojson.go
JivanAmara: :+1: Plus log an error or warning that you've found an invalid polygon & are skipping it.

ARolek:
Thanks for the review. I'm going to push these changes in just a min. 

The only thing I don't like about this approach is the extra memory allocations. Numerous copies are made to just add a single point and Go may take awhile to clean up [large slices](https://github.com/golang/go/issues/14045) the garbage.

One way around this would be to have `closePolygon` accept a `*geom.Polygon` instead of a `geom.Polygoner` and then modify the `*geom.Polygon` by inserting a point when necessary. The method would not need to return anything as the pointer value would be modified. Downside is not accepting the interface, but this is an internal method so that's not really an issue.

ARolek:
here's an example. and the pointer does not need to be passed since it's a slice:

```
func closePolygon(p geom.Polygon) {
	for i := range p {
		//	check if the first point and the last point are the same
		//	if they're not, make a copy of the first point and add it as the last position
		if p[i][0] != p[i][len(p[i])-1] {
			p[i] = append(p[i], p[i][0])
		}
	}
}
```

gdey:
Okay, let me think… I thought I was reusing the space, and only creating a new top-level slice. But the slices at the bottom were just pointed to.

ARolek:
Ok I went ahead an implemented the second strategy. Please review and merge if it looks good.

geom/encoding/geojson/geojson_internal_test.go
```diff
@@ -0,0 +1,118 @@
...
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
+				},
+				{
+					geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9},
+				},
+			},
+			expected: geom.Polygon{
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
```
gdey: Shouldn't this line string have geom.Point{3.2,4.3} at the end as well?

geom/encoding/geojson/geojson_internal_test.go
```diff
@@ -0,0 +1,118 @@
...
+				},
+			},
+			expected: geom.Polygon{
+				{
+					geom.Point{3, 4}, geom.Point{5, 7}, geom.Point{7.6, 8.7}, geom.Point{9.8, 10.9}, geom.Point{3, 4},
+				},
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
```
gdey: Shouldn't this line string have geom.Point{3.2,4.3} at the end as well?

geom/encoding/geojson/geojson_internal_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		"two point only polygon": {
+			geom: geom.Polygon{
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
+				},
+			},
+			expected: geom.Polygon{
+				{
+					geom.Point{3.2, 4.3}, geom.Point{5.4, 6.5},
```
gdey: Shouldn't this line string have geom.Point{3.2,4.3} at the end as well?

geom/encoding/geojson/geojson.go
```diff
@@ -117,3 +129,18 @@ type FeatureCollection struct {
 	Type     featureCollectionType `json:"type"`
 	Features []Feature             `json:"features"`
 }
+
+func closePolygon(p geom.Polygon) {
+	for i := range p {
+		if len(p[i]) < 3 {
+			log.Warn("encounted polygon with less than 3 points")
+			continue
+		}
```
gdey: I'm not sure we should even check for this. I think we should just add the last point. You should check to see if it's as at least one point, otherwise, we have a panic.

ARolek: I was going to initially check for 1 point but decided to check for 3. I'm fine with checking for a single point.

Issue #334, opened 2018-02-24
For invalid x,y values we should return a non-200 response code. (400)


gdey:
The X,Y values should be in the range of:
Each zoom is 2^z tiles wide, so 0 <= x < 2^z and 0 <= y < 2^z

If the x,y value is outside this range, return a 400 error code.

pnorman:
z should also be constrained in a range - at the very least, positive.

gdey:
We already have a check for -z values (https://github.com/terranodo/tegola/blob/v0.6.0/server/handle_map_zxy.go#L47 and https://github.com/terranodo/tegola/blob/v0.6.0/server/handle_map_layer_zxy.go#L49). This should enhance these checks.

ear7h:
@gdey I can take this issue

PR #335, opened 2018-02-25
Refactor X/Y/Z values to uint

### LINKED ISSUE #312: Move z,x,y values to from int to uint
Z/X/Y values should never be negative and therefor should be of type `uint`. This change will need to be implemented in the `server` package for both `HandleMapLayerZXY` and `HandleMapZXY` methods, the `cache` interface and sub packages as well as other packages that accept any z/x/y values (i.e. atlas package: `Map.FilerLayersByZoom()`)

ear7h:
This should close #312 and #334

<s><b>todo</b>
* tile struct in root tile.go</s>


server/handle_map_layer_zxy_test.go
```diff
@@ -73,6 +87,7 @@ func TestHandleMapLayerZXY(t *testing.T) {
 
 		r, err := http.NewRequest(test.reqMethod, test.uri, nil)
 		if err != nil {
+
```
gdey: Let's remove this extra newline.

ARolek:

gdey:
The changes that are needed here is the rename of the import paths.

cache/redis/redis.go
```diff
@@ -71,24 +73,29 @@ func New(config map[string]interface{}) (rcache cache.Interface, err error) {
 		return nil, fmt.Errorf("redis did not respond with 'PONG', '%s'", pong)
 	}
 
+	// the config map's underlying value is int
```
ARolek: If we change this to look for a uint we can drop the less than 0 check on line 82

cache/s3/s3.go
```diff
@@ -62,17 +64,19 @@ func New(config map[string]interface{}) (cache.Interface, error) {
 	//	parse the config
 	c := dict.M(config)
 
-	//	TODO: this could be cleaner
-	defaultMaxZoom := 0
+	// the config map's underlying value is int
+	defaultMaxZoom := tegola.MaxZ
 	maxZoom, err := c.Int(ConfigKeyMaxZoom, &defaultMaxZoom)
```
ARolek: We can change this to parse for a Uint and then drop the less than 0 check on line 74

ARolek:
This commit looks generally fine to me, but we still have a bit of inconsistency in our `uint` strategy. We have a combination of `uint` and `uint64` sprinkled around the codebase all in relation to z/x/y values. We should pick a single way to represent those values and stick with it. I go back and forth on which one to use. `uint` would be nice as it will choose the optimal bit size to use based on the architecture. `uint64` is nice as that's what `strconv` returns when parsing values so it saves on an extra line of code here and there.

I don't have a specific preference on which type we use, we just need to be consistent throughout the codebase when representing Z/X/Y values.

gdey:
I say we just go with uint

cache/s3/s3.go
ear7h: Ah, missed those

ARolek:
LGTM

Issue #204, opened 2017-12-01
Geometry Collection rendering 


CodeKunal:
Are there any plans on supporting rendering of GeometryCollections? As of now it does not support geometry collections. Correct me if I am wrong.

ARolek:
@CodeKunai currently the MVT Specification does not support rendering geometry collections so tegola has no intent to support them unless the spec changes. [Here are the geometry types](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/README.md#434-geometry-types) the MVT spec currently supports.

There is a work arounds you can consider. Check out the PostGIS function [`ST_CollectionExtract`](https://postgis.net/docs/ST_CollectionExtract.html). You could then create three provider layers, one for each type extracted from the Geometry Collection. You would then be able to style points, line strings and polygons independently.

ARolek:
I'm going to close this now as we don't have plans to fix this unless the MVT spec introduces support for it.

PR #342, opened 2018-02-26
Reject PostGIS layers with type ST_GeometryCollection

russss:
I note from #204 that this isn't supported, but for some reason the PostGIS driver accepts it, which caused me some confusion.

I'm diffing against the v0.6.0 branch at the moment, so prod me to rebase if that gets released.

### DIFF
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -319,8 +319,6 @@ func (p Provider) layerGeomType(l *Layer) error {
 					l.geomType = geom.MultiLineString{}
 				case "ST_MultiPolygon":
 					l.geomType = geom.MultiPolygon{}
-				case "ST_GeometryCollection":
-					l.geomType = geom.Collection{}
 				default:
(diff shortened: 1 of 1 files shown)


ARolek:

gdey:
@russss We are trying to make it so that the providers can be shared by other projects. Even though MVT does not support collection types others may. I think we should do the rejection higher up in the stack closer to rendering the MVT layer; as it does not support collections. We are trying to make the entire stack work will collections up to the render point.

ARolek:
I agree with @gdey, this should be handled at the Encoder level as other Encoders (i.e. GeoJSON) do support GeometryCollections. the `mvt` encoder is currently handling this [here](https://github.com/go-spatial/tegola/blob/master/mvt/feature.go#L603). 

Generally speaking the data providers should be responsible for returning all geometries based on the provided query constraints (i.e. zoom, bounding box, user defined SQL) and should not consider downstream capabilities.

russss:
OK - I may have time to take a look at validating further up the chain soon. I also have some more opinions on layer error cases which probably deserve their own ticket...

PR #344, opened 2018-02-27
simplify sort+unique functions

paulmach:
previously, the "find unique" part of these functions would shift everything down when it found unique value(s). This has a worst case runtime of O(n^2). New approach shifts down as you go, O(n). 

While this will probably have little impact on real world performance I found the old version difficult to follow with the 2 extra if conditions.

benchmarks are with 400ish floats, the sort is included as part of the benchmarks:

```
benchmark                              old ns/op     new ns/op     delta
BenchmarkSortUniqueF64-4               16916         15703         -7.17%
BenchmarkSortUniqueF64_worstCase-4     20154         14002         -30.52%
```

### DIFF
--- a/maths/makevalid/makevalid.go
+++ b/maths/makevalid/makevalid.go
@@ -9,10 +9,12 @@ import (
 	"github.com/terranodo/tegola/maths/points"
 )
 
-func allCoordForPts(idx int, pts ...[2]float64) (fs []float64) {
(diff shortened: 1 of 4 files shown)


gdey:
Thanks for the pull request. I will review it soon. Any improvements to the algo speed will help, as this function is called on every feature.

Issue #304, opened 2018-02-15
Figure out why QGIS Vector Tiles Reader compatibility plugin is not able to read our tilejson url.


PetersonGIS:
Add the following end point:
`/capabilities/$map/metadata.json`

Where the json returned follows the follow spec:
https://github.com/geometalab/Vector-Tiles-Reader-QGIS-Plugin/wiki/Help#metadatajson

This will allow QGIS to access tiles from Tegola directly.

gdey:
Would this be in the cache directory? Or at a particular URL? I guess how do you envision this being used?

PetersonGIS:
This might not be an answer, but here's what I was trying to do:
![pluginscreenshot](https://user-images.githubusercontent.com/5266205/36270265-7a5f3b3e-1238-11e8-90fb-b507f49a8bff.JPG)
After clicking save, then connect, it shows this:
![connect](https://user-images.githubusercontent.com/5266205/36270280-86888bb8-1238-11e8-802a-14149f977145.JPG)
I have also tried with this url: https://osm.tegola.io/maps/osm/{z}/{x}/{y}.pbf.
Eventually, if we could get the plugin to work with our tiles and styles then it will help us transform our styles for qgis much more quickly. I imagine there will be some things that will still need to be hand-styled in qgis but not as many as if we started from scratch.

ear7h:
From the spec, it's my understanding that the endpoint should be `/maps/:map/metadata.json` for the metadata since the tiles are being served from `/maps/:map/:z/:x/:y.pbf` and the plugin assumes a specific file structure.

ear7h:
@PetersonGIS After looking into this, I believe the endpoint you should use is `/capabilities/:map` without the trailing `metadata.json`. Let me know if this still doesn't work.

ear7h:
I just checked this against `https://osm.tegola.io/capabilities/osm` and it gave a `Loading error: SSL handshake failed`. So that is something new to look into...

PetersonGIS:
Thanks for the tips @ear7h . I finally got the QGIS Vector Tile Plugin to load up some of our vector tiles with the following URL: https://osm.tegola.io/capabilities/osm. I wonder if the server was down when you tried it before. Can you double-check that it is working for you at this time?

ARolek:
@PetersonGIS can you send over a screen shot of your settings? I still get an error message "None" when trying to connect.

PetersonGIS:
[...]

PR #359, opened 2018-03-12
QGIS vector tile compatibility

ear7h:
Should close issue #304

Added an endpoint `/maps/:map_name/metadata.json` handled by `HandleMapCapabilities` as specified by the QGIS vector tile plugin docs [here](https://github.com/geometalab/Vector-Tiles-Reader-QGIS-Plugin/wiki/Help#metadatajson)


ear7h:
Closing this PR as it is not necessary. See #304.

Issue #322, opened 2018-02-23
un-inferable geometry types prevent layer list from loading on tegola index page.


erictheise:
I haven't been paying much attention to the layer list for the `seamark` project. When I last looked, it didn't load anything. I did a fresh build today and a few renderings trickle through but `syslog` gets many errors:

```
Feb 23 02:00:23 ip-172-31-9-101 dhclient[892]: bound to 172.31.9.101 -- renewal in 1507 seconds.
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: causeway_areas. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: communication_area. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: gridiron_points. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: inshore_traffic_zone_points. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: military_area_points. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: protected_area_points. style defintion not generated
Feb 23 02:10:34 ip-172-31-9-101 env[32683]: 2018/02/23 02:10:34 standard.go:28: #0152018-02-23 02:10:34 [INFO] handle_map_style.go:138: unable to infer geometry type for providerLayerName: radar_line. style defintion not generated
[...]

PR #365, opened 2018-03-19
Check for existence of layers before trying to add them to layer list on index page.

erictheise:
Currently any layer defined in `/capabilities` will be added to the layer list on `index.html`. But for extracts or special schema there is no guarantee that there will be data in a layer. No data leads to `unable to infer geometry type` warnings, failed layer creation within tegola, and a broken layer list on the index page.


erictheise:
I specified the issue number correctly in the branch but incorrectly in the commit message. This relates to #322.

gdey:
I manually cherry-picked this into the minor release of v0.6.2 for the class that Eric has tonight.

Issue #214, opened 2017-12-11
Support for cache purging / seeding using a tile list


ARolek:
The `cache purge` and `cache seed` commands should be able to accept a list of tiles in a file where each line has a `z/x/y` entry. Additionally, the commands should be able to watch a directory for new files and read the files as they're created.

pnorman:
Just to note some [information on an expire list](https://github.com/t-rex-tileserver/t-rex/issues/58)

> How does an expiry list look like?

The standard format is `X Y Z` separated by spaces, with each tile separated by newlines.

For what needs to be implemented, I'd look to render_list from renderd

```
$ render_list --help
Usage: render_list [OPTION] ...
  -a, --all            render all tiles in given zoom level range instead of reading from STDIN
  -f, --force          render tiles even if they seem current
  -m, --map=MAP        render tiles in this map (defaults to 'default')
  -l, --max-load=LOAD  sleep if load is this high (defaults to 16)
  -s, --socket=SOCKET  unix domain socket name for contacting renderd
  -n, --num-threads=N the number of parallel request threads (default 1)
  -t, --tile-dir       tile cache directory (defaults to '/var/lib/mod_tile')
  -z, --min-zoom=ZOOM  filter input to only render tiles greater or equal to this zoom level (default is 0)
  -Z, --max-zoom=ZOOM  filter input to only render tiles less than or equal to this zoom level (default is 20)
If you are using --all, you can restrict the tile range by adding these options:
  -x, --min-x=X        minimum X tile coordinate
  -X, --max-x=X        maximum X tile coordinate
  -y, --min-y=Y        minimum Y tile coordinate
  -Y, --max-y=Y        maximum Y tile coordinate
Without --all, send a list of tiles to be rendered from STDIN in the format:
  X Y Z
e.g.
  0 0 1
  0 1 1
  1 0 1
  1 1 1
The above would cause all 4 tiles at zoom 1 to be rendered
```

The difference with renderd is that it has "dirty" tiles, so it can have an outdated tile still in its store.

For an overview of the different methods, it's worth reading https://github.com/openstreetmap/osm2pgsql/issues/709 and https://github.com/openstreetmap/osm2pgsql/pull/747

ARolek:
@pnorman thanks for the follow up on this. I'm actually about to start working on this and your comment spun up requirement to support a tile entry template. The imposm3 tile list is in z/x/y format. I will make sure to support a configurable template.

ear7h:
[...]

PR #371, opened 2018-03-22
Cache seeding/purging command improvements

ear7h:
There are several improvements made to the cache purge/seed command
* reading from a file via `--file={filename}` flag
* support for varying formats via `--format={format}` flag. 
  * The format is a 4 character string where the first character is the delimiter and the last charcaters are xyz in the file's order.
  * Ex. to parse `X Y Z` use `--format=" xyz"`
* code structure improvements
* using a zoom range for single tile updating

Also a new member function on `slippy.Tile`
* `RangeFamilyAt` - in order to access descendants and ancestors of a tile


pnorman:
It feels like this PR had a rebase messed up and is pulling in extra commits.

> support for varying formats via --format={format} flag.

Is there enough need for different formats to warrant this option?

ear7h:
@pnorman you're right only the last 10ish commits are actually for the PR.
> Is there enough need for different formats to warrant this option?
In #214 you and @ARolek describe two different formats for imposm3 and list_rendered.

ear7h:
moved to #371

PR #372, opened 2018-03-23
Cache seeding/purging command improvements

ear7h:
*cleaned up version of #371

There are several improvements made to the cache purge/seed command
* reading from a file via `--file={filename}` flag
* support for varying formats via `--format={format}` flag. 
  * The format is a 4 character string where the first character is the delimiter and the last charcaters are xyz in the file's order.
  * Ex. to parse `X Y Z` use `--format=" xyz"`
* code structure improvements
* using a zoom range for single tile updating

Also a new member function on `slippy.Tile`
* `RangeFamilyAt` - in order to access descendants and ancestors of a tile


ear7h:
# Comments from #371 

> It feels like this PR had a rebase messed up and is pulling in extra commits.
> 
> > support for varying formats via --format={format} flag.
>
> Is there enough need for different formats to warrant this option?
> -@pnorman 

> @pnorman you're right only the last 10ish commits are actually for the PR.
> 
> > Is there enough need for different formats to warrant this option?
> 
> In #214 you and @ARolek describe two different formats for imposm3 and list_rendered.
> -@ear7h

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+
+	ix := 1
+	iy := strings.Index(format, "y") - 1
+	iz := strings.Index(format, "z") - 1
 
-	parts := strings.Split(str, "/")
+	parts := strings.Split(str, format[:1])
+	//parts := strings.Split(str, "/")
 	if len(parts) != 3 {
 		return tile, fmt.Errorf("invalid zxy value (%v). expecting the format z/x/y", str)
```
gdey: We should return the correct format here.
Here is a play that should allow for parsing the format:
https://play.golang.org/p/HB3kz6n1End
assuming the above play:
```
return tile, fmt.Errorf("invalid zxy value(%v). expecting the format %v", str, format)
```

(As indicated above I would move most of the parsing into the format struct as well).

cmd/tegola/cmd/cache.go
```diff
@@ -19,11 +19,16 @@ import (
 	"github.com/go-spatial/tegola/internal/log"
 	"github.com/go-spatial/tegola/maths"
 	"github.com/go-spatial/tegola/provider"
+	"github.com/go-spatial/tegola/geom/slippy"
+	"os"
```
ARolek: group these at the top of the imports with the rest of the std lib imports

cmd/tegola/cmd/cache.go
```diff
@@ -85,54 +91,20 @@ var cacheCmd = &cobra.Command{
...
-				}
-			} else {
-				//	every zoom
-				for i := uint(0); i <= atlas.MaxZoom; i++ {
-					zooms = append(zooms, i)
-				}
-			}
-
-		}
+		tiles := make(chan *slippy.Tile)
```
ARolek: different name? Tiles could reference a lot of things.

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+func parseTileString(format, str string) (*slippy.Tile, error) {
+	var tile *slippy.Tile
+
+	ix := 1
+	iy := strings.Index(format, "y") - 1
+	iz := strings.Index(format, "z") - 1
 
-	parts := strings.Split(str, "/")
+	parts := strings.Split(str, format[:1])
+	//parts := strings.Split(str, "/")
```
ARolek: remove dead comment

cmd/tegola/cmd/cache.go
ARolek: return nil, not tile

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+func purgeWorker(mt MapTile) {
+
+	z, x, y := mt.Tile.ZXY()
+
+	log.Infof("purging map (%v) tile (%v/%v/%v)", mt.MapName, z, x, y)
+
+	//	lookup the Map
+	m, err := atlas.GetMap(mt.MapName)
+	if err != nil {
+		log.Fatalf("error seeding tile (%+v): %v", mt.Tile, err)
```
ARolek: return error instead of kill the program

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+	}
+
+	//	purge the tile
+	ttile := tegola.NewTile(mt.Tile.ZXY())
+	if err = atlas.PurgeMapTile(m, ttile); err != nil {
+		log.Errorf("error purging tile (%+v): %v", mt.Tile, err)
```
ARolek: return error instead of logging info

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+	}
+}
+
+func sendTiles(zooms []uint, c chan *slippy.Tile) {
+	if cacheZXY != "" {
+		// single xyz
+		//	convert the input into a tile
+		tile, err := parseTileString(cacheFormat, cacheZXY)
+		if err != nil {
+			log.Fatal(err)
```
ARolek: return error instead of kill the program

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+				break ZoomLoop
+			}
+		}
+
+		close(c)
+	} else if cacheFile != "" {
+		// read xyz from a file
+		f, err := os.Open(cacheFile)
+		if err != nil {
+			log.Fatal("could not open file")
```
ARolek: return error instead of kill the program. include file name in error

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+		}
+
+		scanner := bufio.NewScanner(f)
+	ScanLoop:
+		for scanner.Scan() {
+			tile, err := parseTileString(cacheFormat, scanner.Text())
+			if err != nil {
+				log.Fatal(err)
```
ARolek: return error instead of kill the program

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+				}
+			}
+		}
+
+		close(c)
+	} else {
+		// bounding box caching
+		boundsParts := strings.Split(cacheBounds, ",")
+		if len(boundsParts) != 4 {
+			log.Fatal("invalid value for bounds. expecting minx, miny, maxx, maxy")
```
ARolek: return error instead of kill the program

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +167,234 @@ var cacheCmd = &cobra.Command{
...
+		}
+
+		var err error
+		bounds := make([]float64, 4)
+
+		for i := range boundsParts {
+			bounds[i], err = strconv.ParseFloat(boundsParts[i], 64)
+			if err != nil {
+				log.Fatalf("invalid value for bounds (%v). must be a float64", boundsParts[i])
```
ARolek: return error instead of kill the program

ear7h:
Made all the requested changes.

@gdey thank you for the play. I implemented it mostly line for line, except for `newFormat()` unexported function. Since the function is being used within the same file, I thought it clearer to write it as one.

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +178,278 @@ var cacheCmd = &cobra.Command{
...
+	ttile := tegola.NewTile(mt.Tile.ZXY())
+	if err = atlas.PurgeMapTile(m, ttile); err != nil {
+		return fmt.Errorf("error purging tile (%+v): %v", mt.Tile, err)
+	}
+
+	return nil
+}
+
+func sendTiles(zooms []uint, c chan *slippy.Tile) error {
+	defer close(c)
```
gdey: I know you were just copying what was already there, but I had some time to look at this function more carefully. 

This function is hard to read for two reasons:

1. The If-Then-Else block is too long and too deeply nested. 
2. At the end of the function, all you end up doing is returning `nil`, which is the reader has to remember what was going on and then scroll down to the end to find that out. This makes things very had to keep in one's head.

I would recommend changing this over to: 

https://play.golang.org/p/E7DPABYoVa4
or
https://play.golang.org/p/93R58CrBauS

I prefer the `switch` as I think it looks cleaner, and find it easier to read. The understanding with a switch statement is that each block is independent.

cmd/tegola/cmd/root.go
```diff
@@ -39,12 +39,14 @@ func init() {
...
-	cacheCmd.Flags().UintVarP(&cacheMinZoom, "minzoom", "", 0, "min zoom to seed cache from")
-	cacheCmd.Flags().UintVarP(&cacheMaxZoom, "maxzoom", "", 0, "max zoom to seed cache to")
+	cacheCmd.Flags().StringVarP(&cacheTileName, "tile-name", "", "", "operate on a single tile formatted according to tile-name-format")
+	cacheCmd.Flags().StringVarP(&cacheTileList, "tile-list", "", "", "path to a file with tile entries separated by newlines and formatted according to tile-name-format")
+	cacheCmd.Flags().IntVarP(&cacheMinZoom, "min-zoom", "", -1, "min zoom to seed cache from")
+	cacheCmd.Flags().IntVarP(&cacheMaxZoom, "max-zoom", "", -1, "max zoom to seed cache to")
 	cacheCmd.Flags().StringVarP(&cacheBounds, "bounds", "", "-180,-85.0511,180,85.0511", "lat / long bounds to seed the cache with in the format: minx, miny, maxx, maxy")
-	cacheCmd.Flags().IntVarP(&cacheConcurrency, "concurrency", "", runtime.NumCPU(), "the amount of concurrency to use. defaults to the number of CPUs on the machine")
+	cacheCmd.Flags().IntVarP(&cacheConcurrency, "concurrency", "", runtime.NumCPU(), "number of workers to use, defaults to the number of CPUs ")
```
ARolek: remove trailing space.

cmd/tegola/cmd/root.go
```diff
@@ -39,12 +39,14 @@ func init() {
...
 	cacheCmd.Flags().BoolVarP(&cacheOverwrite, "overwrite", "", false, "overwrite the cache if a tile already exists")
+	cacheCmd.Flags().StringVarP(&cacheTileNameFormat, "tile-name-format", "", "/zxy", "4 character string where the first character is a non-numeric delimiter followed by \"z\", \"x\" and \"y\" defining the coordinate order")
```
ARolek: could we use single quotes to avoid the escape?

geom/slippy/tile.go
```diff
@@ -153,3 +188,34 @@ func (t *Tile) BufferedExtent() (bufferedExtent *geom.Extent, srid uint64) {
...
+	}
+
+	// handle descendants
+	mag := zoom - t.z
+	delta := uint(maths.Exp2(uint64(mag)))
+
+	leastX := t.x << mag
+	leastY := t.y << mag
+
+	//log.Info("info: ", mag, delta, leastY, leastY)
```
ARolek: remove or uncomment

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +170,304 @@ var cacheCmd = &cobra.Command{
...
+}
+
+type Format struct {
+	X, Y, Z uint
+	Sep     string
+}
+
+var ErrFormat = errors.New("Invalid format")
+// z/x/y
+var df = Format{
```
ARolek: rename to defaultTileNameFormat or similar

cmd/tegola/cmd/cache.go
```diff
@@ -279,35 +170,304 @@ var cacheCmd = &cobra.Command{
...
+	X:   1,
+	Y:   2,
+	Z:   0,
+	Sep: "/",
+}
+
+func NewFormat(format string) (Format, error) {
```
ARolek: Can we get a test for this function demonstrating single digit and multi digit processing?

ARolek: Scrap the "multi digit process" part of this test request. I misinterpreted what this func was doing. Still could use a test though.

ear7h: I'll add tests

paumas:
When are you planning to merge this pull request?

ARolek:
@paumas plan is to finish testing by the end of the month. Would you be able to help with some testing if we could get you a build?

paumas:
@ARolek yes, I can test. I using dirty tiles list and I can try this seed command with my dirty tiles file.

ARolek:
@paumas beautiful. If you don't mind compiling from source you can pull down this branch and test it. We also have some initial documentation sitting in a PR [here](https://github.com/go-spatial/tegola-docs/pull/7)

ear7h:
@ARolek @gdey This is ready to merge, updated the `vendor.json` with the latest version of the `geom` package

paumas:
@ear7h 
Seed command not working.
I use command: 
`./tegola cache seed --config config.toml --tile-list expired_tiles.txt --overwrite`.
Output console: 
```
2018-05-24 19:28:28 [INFO] config.go:163: loading local config (config.toml)
2018-05-24 19:28:36 [INFO] cache.go:111: zoom list: []
2018-05-24 19:28:36 [INFO] provider.go:95: cleaning up providers
```
My config file:
```
[webserver]
port = ":8001"

[cache]
type = "file"
basepath = "./tegola_cache"

[[providers]]
name = "osm"
type = "postgis"
host = "localhost"
port = 5432
database = "osm"
user = "osm"
password = "osm"
srid = 3857

  [[providers.layers]]
  name = "address"
  geometry_fieldname = "geometry"
  id_fieldname = "id"
  sql = """
    SELECT
      ABS(osm_id) AS id,
      CASE WHEN osm_id < 0 THEN 'r' ELSE 'w' END AS __type__,
      ST_AsBinary(ST_PointOnSurface(way)) AS geometry,
      "addr:housenumber" as housenumber,
      "addr:housename" as name,
      "addr:street" as street,
      "addr:city" as city,
      "addr:postcode" as post_code
    FROM
      planet_osm_polygon
    WHERE
      way && !BBOX! AND
      "addr:housenumber" IS NOT NULL AND "addr:contact" IS NULL
    UNION
    SELECT
      osm_id AS id,
      'n' AS __type__,
      ST_AsBinary(way) as geometry,
      "addr:housenumber" as housenumber,
      "addr:housename" as name,
      "addr:street" as street,
      "addr:city" as city,
      "addr:postcode" as post_code
    FROM
      planet_osm_point
    WHERE
      way && !BBOX!
      AND "addr:housenumber" IS NOT NULL and "addr:contact" IS NULL
  """

[[maps]]
name = "openmap"
attribution = "OpenStreetMap"
center = [23.871, 55.191, 7.0]

  [[maps.layers]]
  name = "address"
  provider_layer = "osm.address"
  min_zoom = 16
  max_zoom = 20
```
My expired tiles file: https://gist.github.com/paumas/b6db154e4dda0c55ec9102551eb82b64

I debug code and I see not working this for: https://github.com/ear7h/tegola/blob/issue-214/cmd/tegola/cmd/cache.go#L130.
My debug code:
```
// range our channel to listen for jobs
println("before");
for mt := range tiler {
	println("seed");
	if gdcmd.IsCancelled() {
		break
	}
	//	we will only have a single command arg so we can switch on index 0
	var err error
	switch args[0] {
	case "seed":
		err = seedWorker(ctx, mt)
	case "purge":
		err = purgeWorker(mt)
	default:
		log.Fatalf("sub-command %q not recognized", args[0])
	}

	if err != nil {
		log.Fatal(err)
	}
}
println("after");
wg.Done()
```
Output console: 
```
2018-05-24 19:44:25 [INFO] config.go:163: loading local config (config.toml)
2018-05-24 19:44:34 [INFO] cache.go:111: zoom list: []
before
before
before
before
before
before
before
before
after
after
after
after
after
after
after
after
2018-05-24 19:44:34 [INFO] provider.go:95: cleaning up providers
```
You can see not output `seed`.

I don't know whats wrong.

paumas:
I found the problem. Problem is I not set `min_zoom` and `max_zoom`.
But, if I use `tile-list`, why I have to set zoom? Zoom exist in the tile-list file.

ear7h:
@paumas Thanks for the feed back. Your diagnosis is correct, this last push should fix that. The specific use case we were looking at provided expired tiles with a single zoom level.

cmd/tegola/cmd/cache.go
```diff
@@ -315,14 +315,19 @@ func sendTiles(zooms []uint, c chan *slippy.Tile) error {
 			}
 			tile := slippy.NewTile(z, x, y, 0, tegola.WebMercator)
 
+			c <- tile
+
 			// range
 			for _, zoom := range zooms {
```
paumas: Why do you need it? You have all the information in the file `cacheFile`.

ear7h: As I said above, one of the use cases ((imposm3)[https://github.com/omniscale/imposm3]) gives a file with uniform zooms. In such a case, the zoom supplied in the file is not really important.

The line you commented ensures the behavior of seeding the tiles directly from the file. This will have to be redesigned...

ear7h:
@paumas The latest push should fix the duplication error.

I believe the fatal error has something to do with file caching on windows, check it here #422

paumas:
@ear7h I use Ubuntu 18.04.
I think problem is duplication problem, one sees worker delete tmp file ant second worker not found tmp file and throw fatal error.
Later, I try run seed command and check your changes.

paumas:
@ear7h I check seed and purge command and everything works perfectly. 
~160.000 tiles seeds ~25 minutes. Good job. Thanks.
@ARolek @gdey I checks this pull-request and this works perfectly. I think we can merge.

paumas:
When are you planning to merge this pull request to master?

agentKfsb:
Все дело не большого короткого времени!))

cmd/tegola/cmd/tile_name_format.go
```diff
@@ -0,0 +1,80 @@
...
+
+var defaultTileNameFormat = Format{
+	X:   0,
+	Y:   1,
+	Z:   2,
+	Sep: "/",
+}
+
+func NewFormat(format string) (Format, error) {
+	// if empty return default
```
gdey: We need to document the format string that is expected by this program:
```go
// NewFormat parses the given format string (in the form of sepxyz, i.e. "/zxy") and returns a new format object.
```

cmd/tegola/cmd/tile_name_format_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		"3 invalid formatStr": {
+			formatStr: "//zxy",
+			err:       fmt.Errorf("invalid formatStr //zxy"),
+		},
+		"4 invalid formatStr": {
+			formatStr: "1zxy",
+			err:       fmt.Errorf("invalid formatStr 1zxy"),
+		},
+	}
```
gdey: Add the following test failure test cases:
"/1xy"
",z45"
"zzxy"
"xyyz
"$xxx"
"$yyy"
"$zzz"
"$xyx"
"$xyy"
"$xzz"
"$xzx"
"$yxx"
"$yxy"
"$xzz"
"$zzz"
"$$$$"
",100"

PR #377, opened 2018-03-27
Issue 375 empty layer should return 400

### LINKED ISSUE #375: Setting max zoom for a map
My map needs a max  zoom at which it will no longer serve and cache tiles.

I tried setting a max_zoom on all the layers that are a part of the map, but a request to a higher zoom was still responded to with a 200 status, and it generated a tile in the cache.

gdey:
This fixes #375


server/server.go
```diff
@@ -13,32 +13,29 @@ import (
...
-	//	is larger than MaxTileSize
+	// MaxTileSize is 500k. Currently just throws a warning when tile
+	// is larger than MaxTileSize
 	MaxTileSize = 500000
 )
 
 var (
-	//	set at runtime from main
+	// set at runtime from main
 	Version string
```
ARolek: remove unused version string

server/server.go
```diff
@@ -50,28 +47,49 @@ func Start(port string) *http.Server {
...
+			"GET", hmzxy.Scheme(),
+			CORSHandler(TileCacheHandler(a, hmzxy)),
+		)
+		group.UsingContext().Handler(
+			"OPTIONS", hmzxy.Scheme(),
+			CORSHandler(hmzxy),
+		)
+		group.UsingContext().Handler(
+			"GET", "/maps/:map_name/style.json",
+			CORSHandler(hmzxy),
```
ARolek: incorrect handler.

server/handle_map_zxy.go
```diff
@@ -113,10 +117,17 @@ func (req HandleMapZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	tile := slippy.NewTile(req.z, req.x, req.y, TileBuffer, tegola.WebMercator)
 
-	//	filter down the layers we need for this zoom
+	// filter down the layers we need for this zoom
 	m = m.FilterLayersByZoom(req.z)
```
ear7h: This might be something for another pull request, but there should also be bounds checking for the map xy as well as z at this part of the request fulfillment

server/server.go
```diff
@@ -50,28 +47,49 @@ func Start(port string) *http.Server {
...
+		)
+	}
 
-	//	static convenience routes
-	group.UsingContext().Handler("GET", "/", http.FileServer(assetFS()))
-	group.UsingContext().Handler("GET", "/*path", http.FileServer(assetFS()))
+	{ // static convenience routes
+		group.UsingContext().Handler("GET", "/", http.FileServer(assetFS()))
+		group.UsingContext().Handler("GET", "/*path", http.FileServer(assetFS()))
+	}
```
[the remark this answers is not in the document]
gdey: @ARolek and I talked going to redo this whole section. I had them in blocks to help separate out the pieces. But, I don't think it helps with readability, as I was hoping.

server/handle_map_zxy.go
```diff
@@ -113,10 +117,17 @@ func (req HandleMapZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	//	check for the debug query string
+	// We filtered out all the layers.
+	if len(m.Layers) == 0 {
+		log.Infof("map (%v) has no layers, at zoom %v", req.mapName, req.z)
+		http.Error(w, fmt.Sprintf("map (%v) has no layers, at zoom %v", req.mapName, req.z), http.StatusBadRequest)
```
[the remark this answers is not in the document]
gdey: Yeah, I'm not sure. Here, seems like @pnorman thinks it should be 400, but it is more like a 404, but this would mean that if a map or layer name that is requested is not defined those should be 404s as well. In #375 we should discuss this.

server/handle_map_zxy.go
gdey: Yeah, that makes sense. We should factor out [this code](https://github.com/go-spatial/tegola/blob/v0.7.0/cache/cache.go#L60) into a helper class to validate zxy values.

ear7h: Well that is implemented in the parseURI method, but I meant to check if the request is within the bounds of map as specified in the server's `config.toml`

server/server.go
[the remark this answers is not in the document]
ear7h: If that's the case it would make it clearer by putting the other groups of routes in blocks as well

server/handle_map_zxy.go
gdey: That's a good point.

server/handle_map_layer_zxy.go
```diff
@@ -105,18 +108,43 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		log.Errorf(errMsg)
-		http.Error(w, errMsg, http.StatusBadRequest)
+		http.Error(w, errMsg, http.StatusNotFound)
 		return
 	}
 
+	// filter down the layers we need for this zoom
+	m = m.FilterLayersByZoom(req.z)
+	if len(m.Layers) == 0 {
+		log.Infof("map (%v) has no layers, at zoom %v", req.mapName, req.z)
```
ARolek: build the string once for both the log and the error

server/handle_map_layer_zxy.go
```diff
@@ -105,18 +108,43 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, fmt.Sprintf("map (%v) has no layers, at zoom %v", req.mapName, req.z), http.StatusNotFound)
+		return
+	}
+
+	if req.layerName != "" {
+		m = m.FilterLayersByName(req.layerName)
+		if len(m.Layers) == 0 {
+			log.Infof("map (%v) has no layers, for LayerName %v at zoom %v", req.mapName, req.layerName, req.z)
```
ARolek: build the string once for both the log and the error

server/handle_map_layer_zxy_test.go
```diff
@@ -3,139 +3,181 @@ package server_test
...
+	if a == nil {
+		a = newTestMapWithLayers(testLayer1, testLayer2, testLayer3)
+	}
+	w, _, err := doRequest(a, tc.method, tc.uri, nil)
+
+	if w.Code != tc.expectedCode {
+		wbody := strings.TrimSpace(w.Body.String())
+		t.Log("wbody", wbody)
+		t.Errorf("status code, expected %v got %v", tc.expectedCode, w.Code)
+	}
```
ARolek: add a return to stop processing

server/handle_map_layer_zxy_test.go
```diff
@@ -3,139 +3,181 @@ package server_test
...
+	// Only try and decode as string for errors.
+	if len(tc.expectedBody) > 0 && tc.expectedCode >= 400 {
+		wbody := strings.TrimSpace(w.Body.String())
+		if string(tc.expectedBody) != wbody {
+			t.Errorf("body, expected %v got %v", string(tc.expectedBody), wbody)
+		}
```
ARolek: add a return to stop processing

Issue #386, opened 2018-04-01
panic: runtime error: invalid memory address or nil pointer dereference


pnorman:
With bcf4ca5 I get

```
$ ~/go/src/github.com/go-spatial/tegola/cmd/tegola/tegola serve
2018-03-31 18:18:47 [INFO] config.go:203: loading local config (config.toml)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x557046240343]

goroutine 1 [running]:
panic(0x5570469207c0, 0xc4200140d0)
        /usr/lib/go-1.7/src/runtime/panic.go:500 +0x1a1
github.com/go-spatial/tegola/cmd/tegola/cmd.initMaps(0xc4202208c0, 0x1, 0x1, 0xc42025b740, 0x0, 0x0)
        /home/pnorman/go/src/github.com/go-spatial/tegola/cmd/tegola/cmd/root.go:181 +0x6b3
github.com/go-spatial/tegola/cmd/tegola/cmd.initConfig()
        /home/pnorman/go/src/github.com/go-spatial/tegola/cmd/tegola/cmd/root.go:82 +0x17d
github.com/go-spatial/tegola/cmd/tegola/cmd.glob..func3(0x557046fbedc0, 0x5570470348a0, 0x0, 0x0)
        /home/pnorman/go/src/github.com/go-spatial/tegola/cmd/tegola/cmd/server.go:21 +0x45
github.com/go-spatial/tegola/vendor/github.com/spf13/cobra.(*Command).execute(0x557046fbedc0, 0x5570470348a0, 0x0, 0x0, 0x557046fbedc0, 0x5570470348a0)
        /home/pnorman/go/src/github.com/go-spatial/tegola/vendor/github.com/spf13/cobra/command.go:702 +0x4a3
github.com/go-spatial/tegola/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x557046fbeba0, 0x0, 0x557046fbedc0, 0xc420189ef0)
        /home/pnorman/go/src/github.com/go-spatial/tegola/vendor/github.com/spf13/cobra/command.go:783 +0x377
github.com/go-spatial/tegola/vendor/github.com/spf13/cobra.(*Command).Execute(0x557046fbeba0, 0x0, 0x0)
        /home/pnorman/go/src/github.com/go-spatial/tegola/vendor/github.com/spf13/cobra/command.go:736 +0x2b
[...]

PR #387, opened 2018-04-01
Fix nil pointer dereference when using implicit zooms

ear7h:
See #386 

Fix to apply layer MaxZoom and MinZoom to underlying config. Was applying to local variable


ARolek:
@ear7h can you add a test to prove the bug and that it's handled? We like to make sure all bug reports have an associated test to avoid the issue in the future.

ARolek:
LGTM. @pnorman when you have a moment can you give this a test with your problematic config?

pnorman:
> LGTM. @pnorman when you have a moment can you give this a test with your problematic config?

WFM with 0c4c0be

PR #391, opened 2018-04-04
Fixes #367/Sort property names in feature inspector based on conversations in that issue.

### LINKED ISSUE #367: Sort the property names before displaying them in the feature inspector box.
When using the Inspector's Inspect Features tool, the order of fields in the pop-up box randomly changes such that a quick visual comparison as the mouse is moved around becomes difficult. Even when the mouse is over one place, if there are multiple features from the same layer underneath the cursor it will show a different field order (as shown below):
inspect_fields

erictheise:


gdey:
@erictheise Talking to @ARolek it seems this is a deeper bug then first thought. It seems that the order of the layers listed should be based on the order they are listed in the config file. So, really what should be happening is that we load the layer names from the config file into an array, to preserve the order, and change out the use of the maps. 

@ARolek I'm of the opinion that we should open up another ticket for that config preserving order, and have this one just make the order static. This is will help people out till the proper fix is in place. I could be convinced otherwise.

ARolek:
This fix LGTM. What @gdey and I were talking about earlier was the order of the map layers, not the attributes in the feature inspector.

Issue #353, opened 2018-03-06
config: environment variable sniffing matches strings with $ not at beginning


ARolek:
A string like `foo = fdsjkl$Gfsdoi423kjfdfX$2Wdsa$dsfdsS$` with match `$G` with our current regex expression when it should be ignored.

gdey:
I think we should get rid of the `$Foo` support and just do `${FOO}` that will make it easier to avoid these types of issues. Not perfect, but Will be better.

ARolek:
This particular issue is because we're not sniffing for beginning of the string using the `${` would be an easier fix. Alternatively we can revisit looking into the `toml.Unmarshaller` interface. If we leverage the interface we will only need to sniff the value, not the key and the value. Downside is we need to wrap every type in our config to do this.

gdey:
That is true, but in this case, where it was a password, it is possible that the password could have started with a `$` then it would not matter; it is less likely that a password starts with a `\${` and ends somewhere down the line with a `}\n`, still possible but less likely. 

Here is a play with my suggestion.

sam-radiant:
Any update on this?

ARolek:
@sam-radiant not quite yet. It's a priority to address though. We're likely going to drop env var support without braces. All env vars will need to look like `${MY_ENV_VAR}`.  I'm hoping we can make some progress on this starting next week.

sam-radiant:
@ARolek Sounds good. I have a VERY temporary workaround. I'll make sure to only use ${} vars as well. Thank you!

ARolek:
@sam-radiant just a heads up, we're almost done with this one. should hit the `v0.7.0` branch this week.

ARolek:
@sam-radiant this has been merged into the v0.7.0 branch and will be part of the next release.

PR #394, opened 2018-04-07
Issue 353 redo env variable support in config package

ear7h:
#353 These changes implement the `toml.UnmarshalTOML` interface in wrapper types for primitives.

One thing to note is that environment variables must be wrapped in quotes.
Ex:

```toml
[nums]
integer1 = "${ENV_SOME_INT}"  # works
integer2 = ${ENV_SOME_INT}     # syntax error
```


config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(varName) == 0 {
+		return in, nil
+	}
+
+	// trim the leading '${' and trailing '}'
+	varName = varName[2:len(varName)-1]
+
+	// get env var
+	envVar := os.Getenv(varName)
+	if envVar == "" {
```
gdey: use [os.LookupEnv](https://godoc.org/os#LookupEnv) here instead of `os.GetEnv`. You should only return this error if the environment var was not defined. Defining it to an empty value may be correct.

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+var EnvironmentError = errors.New("environment variable not found")
+
+var regex *regexp.Regexp
+
+func init() { regex = regexp.MustCompile(`\${[A-Z]+[A-Z1-9_]*}`) }
+
+func replaceEnvVar(in string) (string, error) {
+
+	// TODO(ear7h): FindAllString
+	varName := regex.FindString(in)
```
gdey: You should be getting all the possible replacements; here. This would be useful in a bind like config option, where the user might put something like: `${PG_HOST}:${PG_PORT}`

I would use the [FindAllStringIndex](https://godoc.org/regexp#Regexp.FindAllStringIndex) function. That way you know the index of the places you need to substitute each of the parts.

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
+package env
+
+import (
+	"errors"
+	"os"
+	"regexp"
+	"strconv"
+)
+
+var TypeError = errors.New("type could not be converted")
```
ARolek: This error could communicate more to help users debug.  i.e. "expected type bool, received type 'foo' with value of 'bar'"

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+	var err error
+
+	switch val := v.(type) {
+	case string:
+		val, err = replaceEnvVar(val)
+		if err != nil {
+			break
+		}
+		uintVal, err = strconv.ParseUint(val, 10, 64)
+	case int:
```
ARolek: Our config format currently only needs `uint` support, but I don't think we should be conflating `int` and `uint` here. We should support `int` as a separate type if necessary.

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (t *Bool) UnmarshalTOML(v interface{}) error {
+	var boolVal bool
+	var err error
+
+	switch val := v.(type) {
+	case string:
+		val, err = replaceEnvVar(val)
+		if err != nil {
+			break
```
ARolek: I understand that you're catching the error farther down and returning, but I find this approach a bit confusing to follow. I think if we hav an error we should return then. I know it's extra code but it keeps things consistent.

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+	"regexp"
+	"strconv"
+)
+
+var TypeError = errors.New("type could not be converted")
+var EnvironmentError = errors.New("environment variable not found")
+
+var regex *regexp.Regexp
+
+func init() { regex = regexp.MustCompile(`\${[A-Z]+[A-Z1-9_]*}`) }
```
ARolek: let's add a comment around this regex explaining the string it matches to help future consumers of this package. Also does this need to be in an `init()` method, or can the package var `regex` be set directly?

config/env/env_types.go
```diff
@@ -0,0 +1,161 @@
...
+
+import (
+	"errors"
+	"os"
+	"regexp"
+	"strconv"
+)
+
+var TypeError = errors.New("type could not be converted")
+var EnvironmentError = errors.New("environment variable not found")
```
ARolek: This error could communicate more to help the user. i.e. "environment variable 'FOO' not set"

config/config.go
```diff
@@ -219,10 +220,10 @@ func Load(location string) (conf Config, err error) {
...
 	}
 
-	reader, err = replaceEnvVars(reader)
-	if err != nil {
-		return conf, err
-	}
+	//reader, err = replaceEnvVars(reader)
+	//if err != nil {
+	//	return conf, err
+	//}
```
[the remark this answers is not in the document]
ARolek: Also the old replaceEnvVars code should be  removed with this PR.

config/env/types.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+// matches a variable surrounded by curly braces with leading dollar sign.
+// ex: ${MY_VAR}
+var regex = regexp.MustCompile(`\${[A-Z]+[A-Z1-9_]*}`)
+
+// EnvironmentError corresponds with a missing environment variable
+type EnvironmentError struct {
+	varName string
+}
```
gdey: Not, something I'm asking to change; there is no memory impact. It's just more stylistic.  I would have just used the type here, and not the struct. 
So, for the EnvironmentError:
```
type EnvironmentError string

func (ee EnvironmentError) Error() string {
         return "environment variable "+string(ee)+" not found"
}
```

I only use a struct if there is going to be more items for the error.

PR #395, opened 2018-04-09
vendor clean up and new build flags

ARolek:
- removed several vendored packages
- moved p and cmd packages into the project internal package
- added build flags to remove redis and s3 cache backends


atlas/cache_redis.go
```diff
@@ -0,0 +1,11 @@
+// +build !noRedisCache
+package atlas
```
gdey: I think you need a newline here. 
```
// +build !noRedisCache

package atlas
```

atlas/cache_s3.go
```diff
@@ -0,0 +1,11 @@
+// +build !noS3Cache
+package atlas
```
gdey: I think you need a newline here. 
```
// +build !noS3Cache

package atlas
```

PR #404, opened 2018-04-24
[WIP] Filterer interface

JivanAmara:
Here's what I'm thinking for the Filterer interface.  It addresses a few issues:
* Limit results based on time values and index values in addition to geometrical extent.
* Provide a modification tag to support caching (such as ETag use in wfs3).
* Rename feature Tags -> Properties, It's just a better name :grinning:


gdey:
I don't like how so many concepts are conflated togethering when filtering. Filtering should be considered a separate concept on to its own. I'm not sure about how to deal with timestamps; I keep going back and forth.

Where is what I propose building on what you have right now. Not completely sure on the design.

https://goplay.space/#n3oBVsx3R8z

**File features.go**
   Maybe the concrete filter types should go in there own package? (`filter.Time` or `filter.Extent`)
```go
type TimePeriod struct {
	STime time.Time
	ETime time.Time
}

func (tp *TimePeriod) Start() {
	if tp == nil {
		return time.Time
	}
	return tp.STime
}
func (tp *TimePeriod) End() {
	if tp == nil {
		return time.Time
	}
	return tp.ETime
}

func (_ *TimePeriod) Type() string { return "time" }

type Extent struct {
           *geom.Extent
}
func (_ *Extent) Type() string { return "extent" }

type Properties map[string]interface{}
func (_ Properties) Type() string { return "properties" }

type Feature struct {
	ID         uint64
	Geometry   geom.Geometry
	SRID       uint64
	Properties map[string]interface{}

	// Time value for features with time data.
	// A timestamp would be represented by identical values.
	// This will be nil if the feature does not have time data associated with it.
	Time *TimePeriod

	// This can be generated in a number of ways, for example; hashing the underlying data,
	//	from a modification timestamp, or from a version number.  For read-only data,
	//	a concatenation of the table name & string representation of the feature's ID.
	//	If the provider doesn't support or can't provide this value, it can set this to nil.
	// It needs to be reasonably unique across features in the provider (like the probability of a hash collision)
	// It needs to remain the same if the underlying data doesn't change
	// It needs to change if the underying data changes
	ModificationTag *string
}
```

**file filter.go**
```go
type Filter interface {
	// The type of filter this is
	Type() string
}

type FilterTimer interface {
	Filter
	// use the isZero() method on time to
	// know if there is a start or end time.
	Start() time.Time
	End() time.Time
}

type FilterExtent interface {
	Filter
	Extent() [4]float64
}

type FilterIndex interface {
	Filter
	Start() uint
	End() uint
}

type FilterProperies interface {
	Filter
	Properties() map[string]string
}
```

**file provider.go**
```go
type FeatureConsumer func(f *Feature) error

type Filterer interface {
	StreamFeature(
		ctx context.Context,
		layer string,
		fn FeatureConsumer,
		// Unsupported Filters will be Disgarded.
		filters ...Filter,
	) error
	// Returns the Filters that This Filter supports.
	SupportedFilters() []string
}
```

gdey:
https://goplay.space/#XnA5Af5IpU7

Another thing we can add is:
```go
// ApplyableFilterer can provide a fallback method for providers that don't support them. This
// method is going to be slower as it will do the filter in code.
type ApplyableFilterer interface {
	Filterer
	// Apply allows one to run the filter without the underlining system supporting the filter. This does the filtering in code. If the system return accept, the feature should
	// be passed to the next consumer.
	Apply(f *Feature) (accept bool, err error)
}
```

This will allow filters to provide a fallback method.

JivanAmara:
LGTM.

JivanAmara:
Checks won't pass, as provider.Feature & provider.LayerInfo have changed but not dependent code.  Waiting for consensus on interface & struct changes before updating code that uses them.

JivanAmara:
We need to also update the interface to pass a zoom level in to the provider.  Any objection to adding it as another filter?  It does fit the same concept, though it is used a bit differently behind the scenes & can result in not only filtered results, but different results.

JivanAmara:
The latest two commits simply roll out the basic changes that affect existing code.

ARolek:
@gdey I'm going to close this PR for now. We can revisit it later.

PR #408, opened 2018-05-07
Issue 388 - AWS lambda support

ARolek:


cmd/internal/register/maps.go
```diff
@@ -0,0 +1,170 @@
...
+				DontSimplify:      bool(l.DontSimplify),
+			})
+		}
+
+		// register the map. if an atlas is provided then use it, else use the default atlas
+		if a != nil {
+			a.AddMap(newMap)
+		} else {
+			atlas.AddMap(newMap)
+		}
```
gdey: a `nil` atlas will already add it to the [defaultAtlas](https://github.com/go-spatial/tegola/blob/6b0e9ce614c363073805fa8096079754a21afba0/atlas/atlas.go#L153), so this check is not needed.

cmd/internal/register/maps.go
```diff
@@ -0,0 +1,170 @@
...
+
+			// confirm our providerLayer name is registered
+			var found bool
+			var layerGeomType tegola.Geometry
+			for i := range layerInfos {
+				if layerInfos[i].Name() == providerLayer[1] {
+					found = true
+
+					// read the layerGeomType
+					layerGeomType = layerInfos[i].GeomType()
```
gdey: shouldn't we break here? 

```go
	for i := range layerInfos {
		if layerInfos[i].Name() == providerLayer[1] {
			found = true
			layerGeomType = layerInfos[i].GeoType()
			break
		}
	}

```

cmd/tegola_lambda/main.go
```diff
@@ -0,0 +1,121 @@
...
+
+func main() {
+	var err error
+
+	confLocation := "config.toml"
+
+	// check if the env TEGOLA_CONFIG is set
+	if os.Getenv("TEGOLA_CONFIG") != "" {
+		confLocation = os.Getenv("TEGOLA_CONFIG")
+	}
```
gdey: This should be available on the mail tegola app as well.

PR #409, opened 2018-05-08
Recursive environment variable support

ear7h:
This creates a new `Dicter` interface which abstracts a `map[string]interface{}`. The underlying type can massage the map values as directed by the user. This type enables the config to store something like `map[string]interface{}` with it's own environment variable parsing capabilities.

Dicter:
```go
type Dicter interface {
	String(key string, Default *string) (string, error)
	StringSlice(key string) ([]string, error)

	Bool(key string, Default *bool) (bool, error)
	BoolSlice(key string) ([]bool, error)

	Int(key string, Default *int) (int, error)
	IntSlice(key string) ([]int, error)

	Uint(key string, Default *uint) (uint, error)
	UintSlice(key string) ([]uint, error)

	Float(key string, Default *float64) (float64, error)
	FloatSlice(key string) ([]float64, error)

	Map(key string) (Dicter, error)
	MapSlice(key string) ([]Dicter, error)

	Interface(key string) (v interface{}, ok bool)
}
```


gdey:
@ear7h Seems like one of the moves for the env package did not make it over.

ARolek:
Closing in favor of #411

PR #411, opened 2018-05-10
Ear7h issue 403 "Recursive environment variable support"

ARolek:
This is a copy of @ear7h PR which brings in "Recursive environment variable support" but with fixes to align it with several changes which were recently pushed into the v0.7.0 branch.


cache/redis/redis_test.go
```diff
@@ -32,50 +55,59 @@ func TestNew(t *testing.T) {
...
 			},
-			errMatch: "connection refused",
+			expectedErr: &net.OpError{
+				Op:  "dial",
+				Net: "tcp",
+				Addr: &net.TCPAddr{
+					IP:   net.ParseIP("127.0.0.1"),
+					Port: 6000,
+				},
+				Err: syscall.ECONNREFUSED,
```
ear7h: I changed this to use the the variable defined in the `syscall` package

cache/redis/redis_test.go
```diff
@@ -32,50 +55,59 @@ func TestNew(t *testing.T) {
...
-				continue
-			}
-			t.Errorf("[%v] unexpected err, expected to find %q in %q", i, tc.errMatch, err)
-			continue
-		}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			fn(t, tc)
+		})
```
ear7h: Thanks for cleaning this up

internal/env/map.go
```diff
@@ -0,0 +1,320 @@
...
+package env
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/go-spatial/tegola/internal/dict"
+)
+
+type Map map[string]interface{}
```
ear7h: Should be changed to `env.Dict`. `map` is inconsistent with the other similar package and I think the word "map" could become ambiguous in the presence of geographical maps.

ear7h:
@ARolek LGTM, thank you for the fixes

ear7h:
@ARolek Those are good points, I think it would be the most sane to implement that sort of parsing. Breaking apart the `Dicter` would make things messy when it comes to actually making use of it. For example, the `provider.Init` should't have have logic around the capabilities of a given map-like implementation. I can add this sort of parsing in a separate PR.

Also, I recall that we also need to put guards around the int <-> uint casting.

JivanAmara:
The `Dicter` interface can't go in `internal/` since we plan to break the providers out at some point.  Should we use a dicter for `NewProvider()` calls or leave them as `map[string]interface{}`?

ear7h:
@JivanAmara the plan is to break env and dict into a new repo also. I think this pr should go into v0.7.0 and in the next release we can take all these packages out

gdey:
We should leave it where it is. When we break out the provider packages we can decide then what to do. I'm not sure the provider or the Dicter is in a state currently for public consumption.

internal/env/dict_test.go
```diff
@@ -128,6 +163,31 @@ func TestDict(t *testing.T) {
...
+			dict: env.Dict{
+				"bool_slice": "${TEST_BOOL}",
+			},
+			envVars: map[string]string{
+				"TEST_BOOL": "true, false",
+			},
+			key:      "bool_slice",
+			expected: []bool{true, false},
+		},
+		"bool slice no env": {
```
ear7h: Add test case with `[]bool` in dict

internal/env/parse.go
```diff
@@ -126,3 +201,26 @@ func ParseFloat(v interface{}) (*float64, error) {
...
 }
+
+func ParseFloatSlice(val string) ([]float64, error) {
+	// replace the env vars
+	str, err := replaceEnvVar(val)
+	if err != nil {
+		return nil, err
+	}
+
+	var floats []float64
```
ear7h: Preallocate slice, `Ts := make([]T, 0, len(vals))`

internal/env/parse.go
```diff
@@ -103,6 +155,29 @@ func ParseUint(v interface{}) (*uint, error) {
...
 }
 
+func ParseUintSlice(val string) ([]uint, error) {
+	// replace the env vars
+	str, err := replaceEnvVar(val)
+	if err != nil {
+		return nil, err
+	}
+
+	var uints []uint
```
ear7h: Preallocate slice, `Ts := make([]T, 0, len(vals))`

internal/env/parse.go
```diff
@@ -72,6 +107,29 @@ func ParseInt(v interface{}) (*int, error) {
...
 }
 
+func ParseIntSlice(val string) ([]int, error) {
+	// replace the env vars
+	str, err := replaceEnvVar(val)
+	if err != nil {
+		return nil, err
+	}
+
+	var ints []int
```
ear7h: Preallocate slice, `Ts := make([]T, 0, len(vals))`

internal/env/parse.go
```diff
@@ -42,6 +60,29 @@ func ParseBool(v interface{}) (*bool, error) {
...
 }
 
+func ParseBoolSlice(val string) ([]bool, error) {
+	// replace the env vars
+	str, err := replaceEnvVar(val)
+	if err != nil {
+		return nil, err
+	}
+
+	var bools []bool
```
ear7h: Preallocate slice, `Ts := make([]T, 0, len(vals))`

internal/env/parse.go
```diff
@@ -21,6 +22,23 @@ func ParseString(v interface{}) (*string, error) {
...
 }
 
+func ParseStringSlice(val string) ([]string, error) {
+	// replace the env vars
+	str, err := replaceEnvVar(val)
+	if err != nil {
+		return nil, err
+	}
+
+	var vals []string
```
ear7h: Preallocate slice, `Ts := make([]T, len(vals))`

internal/env/dict.go
```diff
@@ -240,12 +282,26 @@ func (d Dict) Uint(key string, def *uint) (v uint, err error) {
...
+		v, err = ParseUintSlice(val.(string))
+		if err != nil {
+			switch err.(type) {
+			case ErrEnvVar:
+				return v, err
+			default:
+				return v, dict.ErrKeyType{Key: key, Value: val, T: reflect.TypeOf(v)}
+			}
+		}
+	case []int:
```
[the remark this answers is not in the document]
ARolek: yep, good catch.

Issue #413, opened 2018-05-14
Support for S3-like cache


stvno:
I'd like to use Tegola and store the cache on Digital Ocean's 'Spaces'; an S3 like blobstore. The API is similar to S3, but obviously the endpoints are different. Digital Ocean provides an explanation on how to use the open source s3tools to upload data: https://www.digitalocean.com/community/tutorials/how-to-configure-s3cmd-2-x-to-manage-digitalocean-spaces
Using s3cmd it would be possible to use Tegola with more blobstores than just S3

PR #414, opened 2018-05-14
#413 add support for non AWS S3 endpoints

stvno:
Support for non AWS endpoints, based on this example https://www.digitalocean.com/community/questions/how-to-use-digitalocean-spaces-with-the-aws-s3-sdks?answer=39593

This is the first time I've written in Go, so please let me know if I made a mistake

### DIFF
--- a/cache/s3/s3.go
+++ b/cache/s3/s3.go
@@ -31,12 +31,17 @@ const (
 	ConfigKeyBasepath       = "basepath"
 	ConfigKeyMaxZoom        = "max_zoom"
 	ConfigKeyRegion         = "region" // defaults to "us-east-1"
+	ConfigKeyEndpoint       = "endpoint" //	defaults to ""
 	ConfigKeyAWSAccessKeyID = "aws_access_key_id"
 	ConfigKeyAWSSecretKey   = "aws_secret_access_key"
 )
 
 const (
 	DefaultRegion = "us-east-1"
+)
+
+const (
+	DefaultEndpoint = ""
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


gdey:
I'm about to jump on to a plane, but will look at this later today. Have you tested against both aws and digital ocean? Did it work?

stvno:
ah didn't test it with aws I'll test it tomorrow and let you know

stvno:
I just checked and if you do not set an endpoint it works with S3 and if you do set (the correct) endpoint it works with digital ocean :)
However for some reason the tiles are set to private on digital ocean. Which isn't a problem for tegola, since it has a key, but it is for us since we like to serve straight from the blobstore

stvno:
Ah that is also easily remedied, but would require some thought on your side; tegola needs to set the ACL for the tiles and the aws-sdk supports it if you do this:
```
	input := s3.PutObjectInput{
		Body:   aws.ReadSeekCloser(bytes.NewReader(val)),
		Bucket: aws.String(s3c.Bucket),
		Key:    aws.String(k),
		ACL:	aws.String("public-read"),
	}
```
the files are set to public. However I can imagine you want to make this optional or even fully configurable. If you state your preference I can add it to the code if you like. I've created a new issue to track this #415

stvno:
I've added the necessary settings for ACL to this PR

stvno:
Do you need anything else for this PR to be accepted?

gdey:
@stvno Hey sorry, I've had a super busy week last week; I'll try and take a look at it tomorrow.

ARolek:
These changes look good to me less the updates to the package's README. I will accept and then update the readme. Thanks for the contribution!

PR #430, opened 2018-06-13
Issues 429 skip tile feature if geometry nill values

### LINKED ISSUE #429: Empty value stops processing of layer data
It is useful to add some casing to polygon geometries (for example to water features). But as OpenStreetMap water polygons are made of parts (for various reasons), it could give unwanted artefacts as can be seen in this example where smaller river flows into the larger one:
paveikslas
The usual solution is to aggregate geometries with _st_union_ when creating vector tile (aggregating water geometries beforehand for whole region could create huge geometries so it is not a good solution). For example:
```
select
  gid,
  st_union(geom) AS geom
from
  planet_osm_polygon
where
  geom && !BBOX!
```
When a tile has no geometries in a BBOX, this produces a null value for geom (because _st_union_ is an aggregate function). Such null value stops Tegola processing of layer data with following output to console:

`2018/06/09 22:15:25 map.go:193: err fetching tile (z: 10, x: 584, y: 323) features: unable to decode layer (water10) geometry field (geom) into wkb where (gid = 1): EOF`

Now this would not be a problem (if we ignore error message), but if there is another part in a query:
```
select ... st_union(geom) AS geom ...
union all
select somethingelse AS geom from sometable
```
then geometries of the second part of such query are **not** placed into the vector tile basically leaving layer empty in particular tile.
I guess Tegola should handle such _null_ values as "no geometry" and continue with processing query results.

paumas:
Fix #429 issues.
Skip tile feature if geometry nill values.

### DIFF
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -382,6 +382,18 @@ func (p Provider) TileFeatures(ctx context.Context, layer string, tile provider.
 	// fetch rows FieldDescriptions. this gives us the OID for the data types returned to aid in decoding
 	fdescs := rows.FieldDescriptions()
 
+	// find geometry column index
+	geomIndex := -1
+	for i := range fdescs {
+		if fdescs[i].Name == plyr.GeomFieldName() {
+			geomIndex = i
+			break
+		}
+	}
+	if geomIndex == -1 {
+		return fmt.Errorf("error can't find %v field", plyr.GeomFieldName())
+	}
+
 	for rows.Next() {
 		// context check
 		if err := ctx.Err(); err != nil {
@@ -392,6 +404,11 @@ func (p Provider) TileFeatures(ctx context.Context, layer string, tile provider.
 		vals, err := rows.Values()
 		if err != nil {
 			return fmt.Errorf("error running layer (%v) SQL (%v): %v", layer, sql, err)
+		}
+
+		// skip row if geometry nil values.
+		if vals[geomIndex] == nil {
+			continue
 		}
 
 		gid, geobytes, tags, err := decipherFields(ctx, plyr.GeomFieldName(), plyr.IDFieldName(), fdescs, vals)
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -422,14 +422,20 @@ func (p Provider) TileFeatures(ctx context.Context, layer string, tile provider.
 		}
 
 		// decode our WKB
-		geom, err := wkb.DecodeBytes(geobytes)
+		geometry, err := wkb.DecodeBytes(geobytes)
 		if err != nil {
 			return fmt.Errorf("unable to decode layer (%v) geometry field (%v) into wkb where (%v = %v): %v", layer, plyr.GeomFieldName(), plyr.IDFieldName(), gid, err)
 		}
 
+		// skip row if geometry collection empty.
+		g, ok := geometry.(geom.Collection)
+		if ok && len(g.Geometries()) == 0 {
+			continue
+		}
+
 		feature := provider.Feature{
 			ID:       gid,
-			Geometry: geom,
+			Geometry: geometry,
 			SRID:     plyr.SRID(),
 			Tags:     tags,
 		}


ARolek:
@paumas thanks for the PR. I did some looking around and I think we might be able to solve this in a slightly cleaner way. Take a look at the [decipherFields](https://github.com/go-spatial/tegola/blob/v0.7.0/provider/postgis/util.go#L150) function. We check for `nil` values on line 161. We should be able to check the length of `geobytes` after running `decipherFileds` prior to calling `wkb.Decode()` and continue if `len == 0`. What do you think?

We should also put together a test case for this situation.

tomass:
@ARolek was your fix actually implemented?
Tegola fails (Error Getting VTileLayer: Error getting VTileFeature: Unknown geometry type) in the use case https://github.com/go-spatial/tegola/issues/429
I do not see checking of length of geobytes in utils.go
If I apply (merge) @paulmach change - everything works fine.

ARolek:
@tomass the len check was implemented and is currently in master at https://github.com/go-spatial/tegola/blob/master/provider/postgis/postgis.go#L565.

What version of tegola are you using?

I suspect another issue is happening here but I will need to look through @paumas branch to see what functionality it has that the current implementation does not.

tomass:
@ARolek I'm using master with "no_clipping" changes added.
I did compile Tegola with _postgis.go_ version you've given and it is still giving the same error:
`Error: error seeding tile ({z:9 x:286 y:158 Buffer:0 SRID:3857}): Error Getting VTileLayer: Error getting VTileFeature: Unknown geometry type`

@paumas says that the difference is that he is also checking for empty **collection** being returned.

In order to reproduce, you need to have a geometry(ies) which satisfy the condition "&& bbox" but have no intersection with bbox. In my case geometry is like this:

image
here black box is a "bbox" and blue blob is a polygon. Blue polygons bbox crosses the black bbox so feature is selected in where clause but st_intersection selects nothing of it.

Below is an example query to reproduce. In actual production query there is a _!bbox!_ in place of _ST_GeomFromText_:

```
select 1 as gid,
   st_asbinary(st_union(st_intersection(way, ST_GeomFromText('POLYGON((0 0,1 0,0 1,0 0))',3857)))) as geom
from planet_osm_polygon
where way && !bbox!
```

ARolek:
@tomass I looked at the PR again and that's correct, @paumas is sniffing for geometry.Collection inside of PostGIS and ignoring it. I missed that detail on the first review and so this was not implemented.

I don't think we should filter out geometry.Collection at the postgis data provider layer. I think a better approach here would be to ignore the unsupported geometry type during encoding of the mvt feature and log a warning rather than an error. This would give you the same result but I think handled in a more appropriate part of the server. 

@gdey any issues with this?

gdey:
@ARolek I think that is better. An added beneficiate to the approach you are suggesting is that it will apply to all data providers; not just the PostGIS provider.

ARolek:
@tomass & @paumas so I talked with @gdey about this and we're at an interesting fork in the road. I opened up a new issue (#587) which is a higher level discussion around how to we want to handle tile generation errors.

Since it's applicable to the conversation, can you help me understand how your underlying data is using geometry collections? One potential fix is to address the data at the data layer rather than have tegola handle it. This may/may not be possible though, so understanding your situation a bit more would be helpful.

tomass:
@ARolek I'm not sure I understood your question correctly, but here is my opinion.
I would go for skipping the tile entirely when an error occurs. But I would not count empty collection in response as an "error". It is an expected situation.
Here is an example where this situation occurs:
paveikslas
The place is here: https://topo.openmap.lt/#t/14/55.73044/21.09655/0/0
Red lines are tile boundaries. The large waterbody in all tiles except upper right is a Baltic sea plus Curonian lagoon which is a large geometry.
**Firstly I have a cartographic task**: I want all water to have a darker outline, in order to have no lines on water intersection waterbodies must be concatenated before adding them to the tile. I use _st_union_ to do that.
**Secondly I have a technical/speed task**: as the geometry of such a large waterbody is a large one (even tho I do clip it into smaller parts in database beforehand), I do not want to be passing full geometry on all scales to Tegola for clipping, I do the clipping in the PostgreSQL side by using _st_intersection_ and _!bbox!_. And in the place depicted in the picture above we get a situation that bounding box of waterbody overlap with upper right tile - so waterbody geometry matches _where_ clause of the query and is thus selected. But the _st_intersection_ in the _select_ part of the query returns empty geometry (as it is more exact calculation compared to envelope overlapping) which goes into union and thus produces an empty collection.
**Full query**:
```
SELECT
  row_number() over() AS gid,
  ST_AsBinary(ST_Union(ST_Intersection(geom, !BBOX!))) AS geom,
  'coastline' AS kind
FROM
  coastline
WHERE
  geom && !BBOX! AND
  res = 150
```
**Opinion.** Theoretically I could wrap the query in another query and filter out empty geometries, but then there will be other users of Tegola who would hit this situation and it would not be easy for them to understand why are they getting an error of unknown type on **some** tiles or they could fail to see such an error while testing if testing area does not have such a situation.

gdey:
I think we should move some of this conversion to # 587; as that will affect this issue.

PR #444, opened 2018-06-28
Issue 437  Ignore order of points for bounding box caching

### LINKED ISSUE #437: Order of bounds is different in 0.7.0 and master branches
@ear7h,  @ARolek 

https://github.com/go-spatial/tegola/blob/13b50e66d411761df8f4e7d331675c1b4c179621/cmd/tegola/cmd/cache.go#L364

I found that using the same bounds string in 0.7.0 as I did in the master branch for seeding causes no tiles to be cached.  I've tracked it down to around line 366 in the cmd/cache.go file.  

As "up and down" and "top and bottom" can be misinterpreted with different coordinate systems, I would recommend swapping xi and xf if xi > xf (same for the y's).  I think that would help reduce the confusion.

perhaps insert the following around ln 366:

		if xf < xi {
			xi, xf = xf, xi
		}
		if yf < yi {
			yi, yf = yf, yi
		}

I think the purge would need similar logic, if employed.

ear7h:
Should close #437, thank you @TNT0305 for bringing this up and proposing a solution.

### DIFF
--- a/cmd/tegola/cmd/cache.go
+++ b/cmd/tegola/cmd/cache.go
@@ -363,6 +363,7 @@ func sendTiles(zooms []uint, c chan *slippy.Tile) error {
 		_, xi, yi := corner.ZXY()
 		_, xf, yf := diagCorder.ZXY()
 
+		// ensure the initials are smaller than finals
 		if xi > xf {xi, xf = xf, xi}
 		if yi > yi {yi, yf = yf, yi}
 
--- a/cmd/tegola/cmd/cache.go
+++ b/cmd/tegola/cmd/cache.go
@@ -357,11 +357,14 @@ func sendTiles(zooms []uint, c chan *slippy.Tile) error {
 
 		maxZoom := zooms[len(zooms)-1]
 
-		upperLeft := slippy.NewTileLatLon(maxZoom, bounds[1], bounds[0], 0, tegola.WebMercator)
-		bottomRight := slippy.NewTileLatLon(maxZoom, bounds[3], bounds[2], 0, tegola.WebMercator)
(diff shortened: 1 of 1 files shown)


cmd/tegola/cmd/cache.go
```diff
@@ -357,11 +357,15 @@ func sendTiles(zooms []uint, c chan *slippy.Tile) error {
...
+		diagCorder := slippy.NewTileLatLon(maxZoom, bounds[3], bounds[2], 0, tegola.WebMercator)
 
-		_, xi, yi := upperLeft.ZXY()
-		_, xf, yf := bottomRight.ZXY()
+		_, xi, yi := corner.ZXY()
+		_, xf, yf := diagCorder.ZXY()
+
+		// ensure the initials are smaller than finals
+		if xi > xf {xi, xf = xf, xi}
+		if yi > yi {yi, yf = yf, yi}
```
TNT0305: should be:

if yi > yf {yi, yf = yf, yi}

ear7h: oops, thank you for catching that

TNT0305:
yw...stupid fingers!  :p

cmd/tegola/cmd/cache.go
TNT0305: yw...I hate when I do that sort of thing.  I was stuck for 2 hours the other day trying to figure out why the httpPlatformhandler wasn't do *anything*...turned out I had named my config file, "web..config" instead of "web.config".  The font was kinda small and I didn't notice.  DOH!

ARolek:
@ear7h should this PR be closed and not merged pending the work happening on `cache-seed` branch?

Issue #445, opened 2018-07-02
error parsing toml config 


stvno:
I just tried to run the latest version from git and my configuration fails now on: `config: value mapped to "fields" is []interface {} not []string` 

By the looks of it the cause is #411 where `fields, err := vc.StringSlice(ConfigKeyFields)` was changed into `fields, err := layer.StringSlice(ConfigKeyFields)` with `vc` being a dict and `layer` not a dict  in provider/postgis/postgis.go: https://github.com/go-spatial/tegola/blame/638968429bec58da56f1b008ae36d70e207117aa/provider/postgis/postgis.go#L174

Should the config be changed, or is there a bug in the code?

The .config.toml used:
```
[webserver]
port = "0.0.0.0:9797"              # port to bind the web server to. defaults ":8080"

[cache]
type="file"
basepath="/tmp/tegola"

# register data providers
[[providers]]
name = "foo"       # provider name is referenced from map layers (required)
type = "foo"            # the type of data provider. currently only supports postgis (required)
host = "foo"         # postgis database host (required)
port = 5432                 # postgis database port (required)
database = "foo"           # postgis database name (required)
user = "foo"           # postgis database user (required)
password = "foo"               # postgis database password (required)
srid = 3857                 # The default srid for this provider. Defaults to WebMercator (3857) (optiona
max_connections = 50        # The max connections to maintain in the connection pool. Default is 100. (op

# WATER
    [[providers.layers]]
    name = "water1000"
    tablename = "top1000.waterdeel"
    geometry_fieldname = "geometrie_vlak_3857"
    id_fieldname = "ogc_fid"
    srid = 3857
    fields = [ "ogc_fid"]  
```

ARolek:
@stvno  thanks for the report. We will dig into this quickly. The config parsing was overhauled in v0.7.0 and we want to get this fixed before the release. Thanks for testing this. 

@ear7h can you take a look at this?

ear7h:
I can take care of this

stvno:
just to make it clear, I've always used the 0.7.0-branch

ear7h:
The changes were merged into v0.7.0 recently. I just submitted a PR for this issue, fee free to pull it and test it

ARolek:
This is now fixed in v0.7.0

PR #446, opened 2018-07-03
Issue 445 env package not properly parsing slices

ear7h:
For my explanation, the letter "T" represents all types ("string", "bool", "int", etc.)

The error was a result of passing the `[]interface{}` into `ParseT()` vs a single element from the `[]interface{}`. Also added tests for converting from `[]interface{}`to `[]T` and `interface{}` to `T`


ear7h:
Just to mention this should close #445

ARolek:
LGTM

Issue #447, opened 2018-07-03
Set Cache-Control headers on S3 cache


stvno:
It would be nice if Tegola would add Cache-Control headers to S3-cached tiles. e.g. `Cache-Control:max-age=86400` to make sure the client will reuse tiles from its cache and reduce the bandwidth usage.

see https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#PutObjectInput.SetCacheControl

PR #448, opened 2018-07-03
enable Cache-Control headers on S3 caches

stvno:
Sorry for the slightly messy commit history, but basically it is a round about way to sync my fork with upstream
In the end it is just a small change in s3.go, similar to the ACL functionality. I've tested it with Digital Ocean's spaces and I'm 99% sure it will work with S3 as well, since it is using the S3 API. solves #447


stvno:
I've added the documentation

ARolek:
LGTM.

Issue #453, opened 2018-07-09
seeding with --overwrite keeps writing the same tiles


stvno:
I have tegola setup with a concurrency of 6 and I'm overwriting the existing tiles in the S3 cache with the --overwrite flag since it contained older versions. However looking carefully to what happened I noticed the same tiles being seeded over and over again: e.g. the /8/130/83 tile is seeded already seeded 3 times in the short output snippet alone.
I checked on the server and indeed it keeps replacing the tile. So while I want to overwrite the previous generation of tiles, I do not want to overwrite the tiles from the exact same seeding run.

```
2018-07-09 14:02:08 [INFO] cache.go:241: seeding map (stb) tile (8/130/83) took: 775.925833ms
2018-07-09 14:02:08 [INFO] cache.go:241: seeding map (stb) tile (9/260/166) took: 51.944165ms
2018-07-09 14:02:08 [INFO] cache.go:241: seeding map (stb) tile (10/521/332) took: 129.698663ms
2018-07-09 14:02:09 [INFO] cache.go:241: seeding map (stb) tile (6/32/20) took: 2.65048549s
2018-07-09 14:02:09 [INFO] cache.go:241: seeding map (stb) tile (11/1042/665) took: 557.377432ms
2018-07-09 14:02:09 [INFO] cache.go:241: seeding map (stb) tile (13/4171/2662) took: 41.255042ms
2018-07-09 14:02:09 [INFO] cache.go:241: seeding map (stb) tile (14/8343/5325) took: 42.64537ms
2018-07-09 14:02:09 [INFO] cache.go:241: seeding map (stb) tile (15/16686/10650) took: 24.505406ms
2018-07-09 14:02:10 [INFO] cache.go:241: seeding map (stb) tile (12/2085/1331) took: 1.50963916s
2018-07-09 14:02:11 [INFO] cache.go:241: seeding map (stb) tile (7/65/41) took: 4.907601773s
2018-07-09 14:02:12 [INFO] cache.go:241: seeding map (stb) tile (18/133489/85206) took: 5.764526545s
2018-07-09 14:02:12 [INFO] cache.go:241: seeding map (stb) tile (17/66744/42603) took: 5.898816347s
2018-07-09 14:02:12 [INFO] cache.go:241: seeding map (stb) tile (16/33372/21301) took: 6.575422827s
2018-07-09 14:02:12 [INFO] cache.go:241: seeding map (stb) tile (8/130/83) took: 69.644882ms
[...]

PR #454, opened 2018-07-09
Cache seed improvements

### LINKED ISSUE #437: Order of bounds is different in 0.7.0 and master branches

ear7h:
**changes**
* ignore order of bbox coordinates
* allow 0 tile buffer to be used
* test on `sendTiles` function
* test on `SeedTile` function
* corrected default value for `max_zoom`
* memory cache is now registrable 

should close #437 and partially #453


ear7h:
Also, this branch covers the changes in #444

stvno:
Just merged this branch into my fork and the cache-seed performance improvement seems to be huge!

ARolek:
LGTM. only request is to add the other test cases for the SeedTile switch.

PR #455, opened 2018-07-09
Issue 82 SSL support

ear7h:
**changes**
~* most config logic moved to `server.New` function instead of commands~
~* `tile_buffer` now property of `webserver`~
~* `webserver` config uses pointers to place default values~
* ssl support in webserver config
* test for ssl support

~**todo**~
~* tests for https~
~* push middleware~
~* remove globals~


ARolek:
The new config params need to be add to the tegola.io documentation once v0.11.x is released.

PR #461, opened 2018-07-16
postgis: add support for !bbox! (Mapnik) and !BOX! (MapServer) tokens

### LINKED ISSUE #443: !BBOX! token is incompatible with Mapnik
Tegola requires `!BBOX!` (case-sensitive). Mapnik requires `!bbox!` (also case-sensitive).

This makes it inconvenient to copy SQL from style definitions.

olt:
Replacing both variations to !BBOX! on init, to avoid conversion for each request.

Closes #443


ARolek:
LGTM. I changed the base to the v0.7.0 branch as that's what we're working off of right now for features and non critical bugs. Thanks for the PR!

PR #462, opened 2018-07-16
enable Go pprof profiler with TEGOLA_HTTP_PPROF environment

olt:
Allow to use Go's pprof profiler.

This should have no measurable impact when not enabled via ``TEGOLA_HTTP_PPROF``, but I did not found any reliable source for that.

I added a build flag anyway, but can remove it or flip the default (i.e. make it `+build pprof`).


ARolek:
@olt good call on this PR. I think we should flip the build flag to `+build pprof` and allow the bind string to be set via the env var, i.e. `HTTP_TEGOLA_PPROF_BIND=localhost:6060`. I think this would be more explicit and flexible.

ARolek:
@olt that's a good point regarding `LAYER_SQL` and `EXECUTE_SQL`. Those are leftover from the early cuts of tegola. I will open an issue to get those streamlined.

Issue #180, opened 2017-11-13
providers/postgis: remove ST_AsBinary(geom) requirement for sql statements


ARolek:
The `ST_AsBinary(geom)` requirement in SQL statements is a wart that I would like to get rid of. The underlying postgres driver (pgx) supports binary transfer protocol. We should be able to fetch request using the binary protocol rather than the text protocol to remove this wart.

When this task is worked on, make sure to address the outstanding TODO [here](https://github.com/terranodo/tegola/blob/master/provider/postgis/postgis.go#L266).

tomass:
Would it be enough to:
1. Add a new configuration parameter:
`geometry_type=point|linestring|polygon|multipoint|multilinestring|multipolygon`
2. In `tegola/postgis.go` function `layerGeomType` check if geometry_type is specified, if so - use geometry type from there, otherwise - do the "old" logic - replace _st_asbinary_ to _st_geometrytype_ and execute the query.

ARolek:
@tomass 

1. I don't think this will address the topic issue, but this is absolutely something I want to add. This will help when we can't sniff the geom type appropriately. A new issue needs to be opened around this config param addition.
2. Ideally we can drop the ST_AsBinary requirement, I just need to look a bit more into it. I think we need to return everything in binary which should also give us a performance improvement. 

Are you interested in helping with some research?

tomass:
@ARolek yes, I'm interested. @paumas has already done some research on this:
1. Removing (disabling) `layerGeomType` function gives warning in different places where geometry type is used (like displaying list of layers provided and generating _all.json_ info), but vector tiles are rendered just fine.
2. _WKB_ returned by _ST_AsBinary_ already contains geometry type.

ARolek:
@tomass 

1. Correct, the tiles don't need this, it's just used for meta data and generating a style for the built in viewer.
2. Correct, the main goal for dropping the ST_AsBinary requirement is for usability.

If we were to remove auto sniffing the geometry type we could purge a bunch of complex code. It might be best overall to have the user define the geom type in the config file instead of sniffing for it, the downside being more work required by the end user.

tomass:
Code for sniffing is already there. So maybe both methods could be supported? Somebody could then specify geometry type only for advanced usage (like returning WKB pre-calculated in materialized view or similar). Leaving the sniffing code would also remove the necessity to update the config after upgrading Tegola.
But it is up to you to decide.

ARolek:
[...]

PR #466, opened 2018-07-19
provider/postgis: add geometry_type option to avoid table inspection

olt:
addressed in #180 

This feature is important for more complex queries (joins or spatial operations) were a `LIMIT 1` still has to run against the whole table (because !BBOX! covers the whole planet).


ARolek:
@olt this looks great! Also, thanks for cleaning up the internal PostGIS test.

PR #472, opened 2018-07-28
Issue 383 - Id field as tag

### LINKED ISSUE #383: provider/postgis: The `id` field is not included as a property if specified in `fields` in the config
# Context
When explicitly specifying the `id_fieldname` in the `fields` collection of a `config.toml` definition, then the `id` field is not included as a tag as expected. 

Example:
```
    [[providers.layers]]
    name = "geo_place"
    tablename = "metrics.dim_geo_places"
    id_fieldname = "id"
    geometry_fieldname = "shape"
    fields = [
        "id",    <--- This property is ignored...
        "name",
        "level",
        "valid_from",
        "valid_to"
    ]
```

I had a quick look at the source code and found the culprit here:
 - https://github.com/go-spatial/tegola/blob/6a08f2c826fe849dfd4571a58b4361c6c95aa7e0/provider/postgis/util.go#L169

I am no expert in go, but somehow I think that if the `id` field was explicitly specified in the `fields` array, then it should be included as a tag. My pseudo code solution would look something like this:
```
...
case idFieldname:
	gid, err = gId(v)
	if in_collection(descriptions, idFieldname) {
		tags[desc.Name] = value
	}
...
```

For now the work around is to explicitly define an additional column with another name, eg:
```
SELECT 
    gid, 
    ST_AsBinary(geom) as geom,
    id as my_id_column,     <---- this is the workaround...
FROM ...
WHERE geom && !BBOX!
```

Hope this makes sense, and helps.

ear7h:
should close #383 

The sql generator always adds a `gid` field, if config also includes the `gid` field as a tag, it is added twice. Meaning there could be two identical `gid` columns. The `decipherFields` function, also parses the first appearance of `gid` as usual but the second as a tag.

### DIFF
--- a/provider/postgis/util.go
+++ b/provider/postgis/util.go
@@ -49,20 +49,13 @@ func genSQL(l *Layer, pool *pgx.ConnPool, tblname string, flds []string) (sql st
 		}
 	}
 
-	for i := range flds {
-		flds[i] = fmt.Sprintf(`"%v"`, flds[i])
-	}
+	fgeom := -1
 
-	var fgeom int = -1
-	var fgid bool
 	for i, f := range flds {
-		if f == `"`+l.geomField+`"` {
+		if f == l.geomField {
 			fgeom = i
 		}
-
-		if f == `"`+l.idField+`"` {
-			fgid = true
-		}
+		flds[i] = fmt.Sprintf(`"%v"`, flds[i])
 	}
 
 	// to avoid field names possibly colliding with Postgres keywords,
@@ -73,7 +66,8 @@ func genSQL(l *Layer, pool *pgx.ConnPool, tblname string, flds []string) (sql st
 		flds[fgeom] = fmt.Sprintf(`ST_AsBinary("%v") AS "%[1]v"`, l.geomField)
 	}
 
-	if !fgid && l.idField != "" {
+	// add required id field
+	if l.idField != "" {
 		flds = append(flds, fmt.Sprintf(`"%v"`, l.idField))
 	}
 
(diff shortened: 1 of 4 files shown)


gdey:
@ear7h @ARolek This seems like something we should get in as well.

ARolek:
@ear7h are you able to change the base branch on this to v0.11.x? GH is not allowing the change from my side.

ear7h:
@ARolek this pr is ready

provider/postgis/util.go
```diff
@@ -50,20 +50,13 @@ func genSQL(l *Layer, pool *pgx.ConnPool, tblname string, flds []string) (sql st
 		}
 	}
 
-	for i := range flds {
```
ARolek: I think this code was there to wrap all the SELECT values in quotes to avoid issues when the field is also a postgres keyword (i.e. name -> "name"). I don't recall if we have a test for this, but I think we should double-check before we purge this code. Looks like some comments would be helpful here.

provider/postgis/util.go
```diff
@@ -207,14 +202,23 @@ func decipherFields(ctx context.Context, geomFieldname, idFieldname string, desc
 		}
 
 		desc := descriptions[i]
+		// the id has to be parsed once but it can also be a tag
```
ARolek: this comment looks out of place. I see it repeat in line 213.

provider/postgis/util.go
ear7h: The diff is messed up, and it's best seen side-by-side. If you look at line 55 and 59 of the new file the looping and adding quotes is still there. In the old file, it was looping twice and doing checks against the new (quoted) fields

ARolek:
@ear7h please squash then rebase & merge into v0.11.x

PR #477, opened 2018-07-31
provider/postgis: replace Mapnik pixel_width/height and scale_denominator SQL tokens

olt:


ARolek:
@olt I had a few minor changes which I made in PR #503, mainly making the tokens case insensitive so we could stay in alignment with the uppercase tokens already in tegola. Thanks for the PR.

Issue #467, opened 2018-07-19
provider/postgis: Support for sub-queries as "tablename"


olt:
Mapnik only supports table names, but no SELECT statements. Mapnik always adds something like `SELECT geometry FROM ... WHERE ST_Intersects(geometry, !bbox!)` around the table name. Complex queries are supported by wrapping the query in `(...) AS subquery`.

So `(SELECT id, geometry, ST_Area(geometry) as area FROM landuse WHERE ST_Intersects(geometry, !bbox!)) AS sub` is a valid "tablename".

See https://github.com/mapnik/mapnik/wiki/PostGIS

I'd like to either 
- add support for sub-queries in the `tablename` option. Only need to `replaceTokens` before inspecting fields in `genSQL`.
- add support for sub-queries in the `sql` option. Need to inspect the statement to determine if is a `SELECT` or `() AS sub` query and then handle it like a `tablename`.

~~I prefer the first.~~

PR #478, opened 2018-07-31
provider/postgis: support for SQL sub-queries

olt:
Implements #467 and also refactors and extends TestTileFeatures.


ARolek:
@olt I really like this implementation. It's a "clean hack" that is quite effective. I made some minor changes in another branch but overall this is a great approach to adding in subqueries. 👏

ARolek:
The new PR is #507.

PR #486, opened 2018-08-06
Improve GeoPackage compatibility

olt:
Fixes various issues with GeoPackages that do not use fid/geom and do not use quotes around column names.

### DIFF
--- a/provider/gpkg/gpkg_register.go
+++ b/provider/gpkg/gpkg_register.go
@@ -3,6 +3,7 @@
 package gpkg
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


provider/gpkg/gpkg_register.go
```diff
@@ -82,29 +83,74 @@ func AutoConfig(gpkgPath string) (map[string]interface{}, error) {
...
-	colPattern := `"(.+?)"`
-	colFinder := regexp.MustCompile(colPattern)
+	var pkCol string
+	colNames := make([]string, 0, len(defs))
 
-	// Get all column names (drop "CREATE TABLE (" portion)
-	colDefs := colDefsFinder.FindStringSubmatch(sql)[1]
-	colMatches := colFinder.FindAllString(colDefs, -1)
+	// match unquoted (`column_name`) or quoted (`"column name"`) indentifiers
+	colFinder := regexp.MustCompile(`^(([a-zA-Z_][a-zA-Z0-9_]*)|"([^"]+)")\s`)
```
gdey: Would it make sense to have this regex as global var, build up in an init function, so that it does not have to be reinitialized every time this function is called.

ARolek:
LGTM

Issue #500, opened 2018-08-24
cache cmd: lat long bounds cannot cross the anti meridian


ear7h:
The current math refuses to go over the anti meridian. Running the following command:

tegola cache seed --bounds="179.99,0.001,-179.99,0.002" --max-zoom=3 --min-zoom=3
Produces:

2018-08-24 14:41:10 [INFO] config.go:163: loading local config (config.toml)
2018-08-24 14:41:10 [INFO] cache.go:106: zoom list: [3]
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/0/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/1/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/2/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/4/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/5/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/3/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/7/3)
2018-08-24 14:41:10 [INFO] cache.go:232: cache seed set to not overwrite existing tiles. skipping map (zoning) tile (3/6/3)
2018-08-24 14:41:10 [INFO] provider.go:95: cleaning up providers
Which is all the tiles where y=3. The command should only have seeded tiles 3/0/3 and 3/7/3 where the bounds have a minimum (westward) longitude of 179.99 and maximum (eastward) longitude of -179.99.

ear7h:
related to #437

PR #498, opened 2018-08-24
Refactor according to `geom` changes for slippy tiles

ear7h:
The following PR introduced api changes to slippy tiles, this PR updates the `geom` dependency and refactors accordingly.
https://github.com/go-spatial/geom/pull/22

PR #502, opened 2018-08-25
Issue 500 cache bounds

ear7h:
#500


ear7h:
This pr depends on #498

PR #503, opened 2018-09-04
Issue 468

ARolek:


provider/postgis/util.go
```diff
@@ -95,14 +102,50 @@ func replaceTokens(sql string, srid uint64, tile provider.Tile) (string, error)
...
+//	uppercaseTokens will sniff for ! chars and uppercase everything between them
+//	if an odd number of ! are found an error is thrown
+func uppercaseTokens(str string) (string, error) {
+	rs := []rune(str)
+
+	uppercase := false
+	for i := range rs {
+		if rs[i] == '!' {
+			uppercase = !uppercase
+		}
```
gdey: You should have a continue

provider/postgis/util_internal_test.go
```diff
@@ -61,6 +67,52 @@ func TestReplaceTokens(t *testing.T) {
...
+		},
+		"no tokens": {
+			str:      "no token",
+			expected: "no token",
+		},
+		"unclosed token": {
+			str:         "unclosed !token",
+			expectedErr: ErrUnclosedToken("unclosed !token"),
+		},
+	}
```
gdey: Add an empty string test case, not just a no token test case.

PR #508, opened 2018-09-07
Refactor according to `geom` changes for slippy tiles (rehash)

ear7h:
This is a rehash of #498, the changes from that pr became stale and a new pr is cleaner.


gdey:
@ear7h It looks like this pull request is similar to the one you just did and got merged. Can you verify if this is still valid?

ear7h:
@gdey that's correct, closing this. For reference, the pr is #622

PR #509, opened 2018-09-08
Issue 438: gzip compression

ARolek:
@stvno and @pnorman please give this PR a test if you have a chance. 

*IMPORTANT*: if you have a current cache in place, you will need to purge it entirely as tegola now expects the cache to persist gzipped tiles.


stvno:
There is a bug in the implementation of the GZipHandler:
Firefox sends the Accept-Encoding header as "gzip, deflate". The if/else loop structure loops through each of the accepted encodings in stead of stopping after it found a gzip-compatible one. As such it will deflate the result because `deflate` is not `gzip` or `*`

stvno:
The S3 cache doesn't set the Encoding-Type right now, which means that serving directly from S3 doesn't work.
If I add `ContentEncoding: aws.String("gzip"),` to the Set function, Tegola crashes because the it tests the S3 permissions with a non-gzipped object:
`if err := s3cache.Set(&key, []byte("\x53\x69\x6c\x61\x73")); err != nil {` and it expects a gzipped object.

I guess the solution would be to gzip the test object since we gzip everything from now on.

server/middleware_gzip.go
```diff
@@ -0,0 +1,90 @@
...
+					next.ServeHTTP(&gzipDecompressResponseWriter{resp: w}, r)
+					return
+				}
+				break
+			} else {
+				//	decompress
+				next.ServeHTTP(&gzipDecompressResponseWriter{resp: w}, r)
+				return
+			}
+		}
```
gdey:
```go
decompress := false
for _, v := range strings.Split(acceptEncoding, ",") {
         if  (strings.Contains(v,"gzip") || strings.Contains(v,"*")) && !strings.HasSuffix(v,";q=0") {
           decompress = true
        } 
}
if decompress {
        next.ServeHTTP(&gzipDecompressResponseWriter{resp: w}, r)
	return
}
// set appropriate header
w.Header().Set("Content-Encoding", "gzip")
next.ServeHTTP(w, r)
return
```

This is cleaner to read and fixes the issue if the gzip header is the second one. Please, add a test case for this.

server/middleware_gzip_test.go
```diff
@@ -0,0 +1,114 @@
...
+				"Content-Encoding": "gzip",
+			},
+		},
+		"Accept-Encoding: *;q=0": {
+			uri: "/maps/test-map/10/2/3.pbf",
+			requestHeaders: map[string]string{
+				"Accept-Encoding": "*;q=0",
+			},
+			expectedResponseHeaders: map[string]string{},
+		},
```
gdey: Add a test case with the gzip encoding in the second position. Same with a "q=0", version.

PR #510, opened 2018-09-10
fix mix-up in SQL token documentation

olt:


ARolek:
oops. good catch! thank you.

Issue #369, opened 2018-03-22
Set default_transaction_read_only when connecting to PostgreSQL


pnorman:
Yes, we should all be making sure that tegola connects from a database user who only has SELECT access, but how many of us are doing that? ;)

[`default_transaction_read_only`](https://www.postgresql.org/docs/current/static/runtime-config-client.html#GUC-DEFAULT-TRANSACTION-READ-ONLY) makes sure that a style bug can't escalate to altering data in the database

With libpq this could be done with either `SET default_transaction_read_only = TRUE` or setting options=`-c default_transaction_read_only=true`.

With pgx this could either the SQL or setting `RuntimeParams` of the [`ConnConfig`](https://godoc.org/github.com/jackc/pgx#ConnConfig)

c.f. mapnik/mapnik#2142

gdey:
I think this would be a good thing to add.

We would just need to add the option here:
https://github.com/go-spatial/tegola/blob/master/provider/postgis/postgis.go#L141
```go
RuntimeParams: map[string]string{
               "default_transaction_read_only" : "TRUE",
}
```
Postgress documentation:
https://www.postgresql.org/docs/9.2/static/sql-set.html
https://www.postgresql.org/docs/9.2/static/sql-show.html

https://www.postgresql.org/docs/9.2/static/runtime-config.html
https://www.postgresql.org/docs/9.2/static/runtime-config-client.html

PR #512, opened 2018-09-12
Add read only runtimeParm to Postgress

Hoovs:
Created for issue #369


ARolek:
@Hoovs I made a minor change and added an extra runtime param. I'm going to include your commit in another PR. Thanks for the contribution!

PR #516, opened 2018-09-17
Updated docker image

### LINKED ISSUE #385: Add CA certificates to Docker container
The tegola Dockerfile needs CA certificates in order to make calls to SSL enabled endpoints (i.e. fetching a remote config). The following command needs to be added [here](https://github.com/go-spatial/tegola/blob/master/docker/Dockerfile#L52). 

`RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*`

stvno:
working docker for the 0.8.0 branch, should fix #385

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,16 +26,15 @@
 
 
 # --- Build the binary
-FROM golang:1.10.3-alpine3.7 AS build
+FROM golang:1.11.0-alpine3.8 AS build
 
 # Only needed for CGO support at time of build, results in no noticable change in binary size
 #  incurs approximately 1:30 extra build time (1:54 vs 0:27) to install packages.  Doesn't impact
 #  development as these layers are drawn from cache after the first build.
 RUN apk update
-RUN apk add musl-dev=1.1.18-r3
-RUN apk add gcc=6.4.0-r5
+RUN apk add musl-dev=1.1.19-r10
+RUN apk add gcc=6.4.0-r9
 RUN apk add git
-
 
 # Set up source for compilation
 RUN mkdir -p /opt/tegola_src
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -53,8 +53,7 @@ RUN chmod a+x /opt/tegola
 
 # --- Create minimal deployment image, just alpine & the binary
 FROM alpine:3.7
-RUN apk update
-RUN apk add ca-certificates
+RUN apk update && apk add ca-certificates && rm -rf /var/cache/apk/*
 LABEL maintainer="Development@JivanAmara.net"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -26,14 +26,16 @@
 
 
 # --- Build the binary
-FROM golang:1.9.2-alpine3.7 AS build
+FROM golang:1.10.3-alpine3.7 AS build
```
ARolek: Can we bump to Go 1.11 now that it's out?

docker/Dockerfile
```diff
@@ -45,12 +47,14 @@ RUN ln -s /opt/tegola_src /opt/tegola_src/src/github.com/go-spatial/tegola
 # Build binary
 WORKDIR /opt/tegola_src/cmd/tegola
 ENV GOPATH=/opt/tegola_src
+RUN go get -u github.com/theckman/goconstraint/go1.8/gte
```
ARolek: Ideally we should vendor this rather than add it here.

docker/Dockerfile
ARolek: looks like `golang:1.11.0-alpine3.8` is now available

docker/Dockerfile
ARolek: I looked into it and this package is [vendored](https://github.com/go-spatial/tegola/blob/master/vendor/vendor.json#L407). Is the Docker build failing without this line?

stvno: Yes it states:
```
Step 12/19 : RUN go build -v -gcflags "-N -l" -o /opt/tegola
 ---> Running in 662cda042194
main.go:7:2: cannot find package "github.com/theckman/goconstraint/go1.8/gte" in any of:
        /usr/local/go/src/github.com/theckman/goconstraint/go1.8/gte (from $GOROOT)
        /opt/tegola_src/src/github.com/theckman/goconstraint/go1.8/gte (from $GOPATH)
```

docker/Dockerfile
stvno: done

docker/Dockerfile
ARolek: Strange. Can you check your local copy. I see this in the vendor directory [here](https://github.com/go-spatial/tegola/tree/v0.8.x/vendor/github.com/theckman/goconstraint/go1.8/gte) and I would prefer to keep this go get request out of the Docker image.

ARolek:
Looking into this a bit more I think it has to do with the way the Docker image is being built. The GOPATH is being overwritten and the DockerFile is kind of working against the root image. Let me do a bit more research.

gdey:
#535 Redid this commit.

PR #522, opened 2018-10-02
don't require IDs for mvt features. closes #338

### LINKED ISSUE #338: Don't require ID field for layers
MVT doesn't require a ID field. Many sources won't have an ID on all layers, and this is the norm on OSM-derived ocean data.

ARolek:
- removed the unique ID check to match the mvt spec
which says IDs SHOULD be unique but is not a MUST.
- reworked mvt tests to closer align to the new testing
structure and remove the dependency on tbltest package

### DIFF
--- a/mvt/cleanline_internal_test.go
+++ b/mvt/cleanline_internal_test.go
@@ -18,7 +18,7 @@ func TestCleanline(t *testing.T) {
 			el := cleanLine(tc.line)
 
 			if !tegola.IsLineStringEqual(el, tc.expected) {
-				t.Errorf("expected %v, Got %v", tc.line.GoString(), el.GoString())
+				t.Errorf("expected %v, got %v", tc.line.GoString(), el.GoString())
 			}
 		}
 	}
--- a/mvt/cleanline_internal_test.go
+++ b/mvt/cleanline_internal_test.go
@@ -3,18 +3,28 @@ package mvt
 import (
 	"testing"
 
-	"github.com/gdey/tbltest"
 	"github.com/go-spatial/tegola"
 	"github.com/go-spatial/tegola/basic"
 )
 
 func TestCleanline(t *testing.T) {
-	type testcase struct {
+	type tcase struct {
 		line     basic.Line
 		expected basic.Line
 	}
-	tests := tbltest.Cases(
-		testcase{
+
+	fn := func(tc tcase) func(t *testing.T) {
+		return func(t *testing.T) {
+			el := cleanLine(tc.line)
+
+			if !tegola.IsLineStringEqual(el, tc.expected) {
+				t.Errorf("expected %v, Got %v", tc.line.GoString(), el.GoString())
+			}
+		}
+	}
+
+	tests := map[string]tcase{
+		"1": tcase{
 			line: basic.Line{ // basic.Line len(000007) direction(clockwise).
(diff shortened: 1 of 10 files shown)


mvt/cleanline_internal_test.go
```diff
@@ -3,47 +3,54 @@ package mvt
...
 	}
-	tests := tbltest.Cases(
-		testcase{
+
+	fn := func(tc tcase) func(t *testing.T) {
+		return func(t *testing.T) {
+			el := cleanLine(tc.line)
+
+			if !tegola.IsLineStringEqual(el, tc.expected) {
+				t.Errorf("expected %v, Got %v", tc.line.GoString(), el.GoString())
```
gdey: `Got` should be `got`

mvt/feature_internal_test.go
```diff
@@ -75,78 +80,82 @@ func TestEncodeGeometry(t *testing.T) {
...
 
-			if j < len(g) && tcase.egeo[j] != g[j] {
-				t.Errorf("[%v] Geometry at %v, Expected %v Got %v", i, j, tcase.egeo[j], g[j])
-				t.Logf("[%v] Geometry, Expected %v Got %v", i, tcase.egeo, g)
-				break
+	fn := func(tc tcase) func(t *testing.T) {
+		return func(t *testing.T) {
+			g, gtype, err := encodeGeometry(context.Background(), tc.geo, tile, true)
+			if tc.eerr != err {
+				t.Errorf("error, Expected %v Got %v", tc.eerr, err)
```
gdey: Expected => expected
Got => got

mvt/feature_internal_test.go
```diff
@@ -75,78 +80,82 @@ func TestEncodeGeometry(t *testing.T) {
...
+			}
+			if gtype != tc.typ {
+				t.Errorf("geometry type, Expected %v Got %v", tc.typ, gtype)
+			}
+			if len(g) != len(tc.egeo) {
+				t.Errorf("geometry length, Expected %v Got %v ", len(tc.egeo), len(g))
+				t.Logf("Geometries, Expected %v Got %v", tc.egeo, g)
```
gdey: lowercase Geometries, Expected and Got.

mvt/layer.go
```diff
@@ -139,38 +139,20 @@ func (l *Layer) Features() (f []Feature) {
...
-
+// AddFeatures will add one or more Features to the Layer
+// per the spec features SHOULD have unique ids but it's not required
+func (l *Layer) AddFeatures(features ...Feature) {
+	// pre allocate memory
 	b := make([]Feature, len(l.features), len(l.features)+len(features))
+
 	copy(b, l.features)
+
 	l.features = b
```
gdey: I would redo this to the following:

```go
// per-allocate memory
b := make([]Feature, len(l.features)+len(features))
copy(b, l.features)
copy(b[len(l.features):],features)
l.features = b
```

copy is faster then append if you just need to copy over memory.

mvt/layer.go
```diff
@@ -146,9 +146,9 @@ func (l *Layer) AddFeatures(features ...Feature) {
 	b := make([]Feature, len(l.features), len(l.features)+len(features))
```
gdey: The make is wrong.
it should be:
```go
b := make([]Feature, len(l.features)+len(features))
```

PR #528, opened 2018-10-12
provider/postgis: fix uppercaseToken for queries with != etc.

olt:
uppercaseTokens fails if the query contains exclamation marks (e.g. bar != 42).
I replaced it with a regexp that only matches tokens like `!Tok-EnS_01!`.


ARolek:
@olt thanks for the PR! Looks good on this end. I guess we can drop `ErrUnclosedToken` now. I will take care of that. 

Thank you!

Issue #529, opened 2018-10-14
How does cache --tile-list option work?


bagage:
While using file cache system, I tried to use the `--tile-list` option to force cache refresh for a given list.

My file contains:

```
1/0/1
2/1/2
3/3/4
4/7/9
5/15/19
6/30/39
6/31/39
7/61/79
7/62/79
8/123/159
8/124/159
9/247/318
9/248/318
9/247/319
9/248/319
10/495/637
10/496/637
10/495/638
10/496/638
```

And I'm reading that file with `tegola cache seed --overwrite --tile-list $WORK_FILE`. Instead of seeing `seeding map (xxx) tile (z/x/y) took: Xms` 19 times (there are **19 tiles** to refresh), actually tegola refresh all my tiles for all zooms (**7416 tiles**, because I limited it to zoom 7)???

Did I miss something?

stvno:
It is broken, the sendTiles function will parse the tilelist-file and for each tile it will test if that tile is within the given zoom range. However the code will return all tiles within that zoomlevel, instead of just the tiles from the list:

https://github.com/go-spatial/tegola/blob/v0.8.x/cmd/tegola/cmd/cache.go#L331

```
// range
			for _, zoom := range zooms {
				err := tile.RangeFamilyAt(zoom, func(t *slippy.Tile) error {
					if gdcmd.IsCancelled() {
						return fmt.Errorf("cache manipulation interrupted")
					}

					if z1, x1, y1 := t.ZXY(); z != z1 || x != x1 || y != y1 {
						c <- t
					}

					return nil
				})

				// graceful stop if cancelled
				if err != nil {
					return err
				}
			}

```

stvno:
I think it should be:

```
if z1, x1, y1 := t.ZXY(); z == z1 && x == x1 && y == y1 {
						c <- t
					}
```

stvno:
this is a **bad** idea: it will check each tile against all possible tiles in the zoomRange :(

ARolek:
@ear7h can you take a look at this? I’m traveling today so won’t havs a chance.

stvno:
I've removed the whole tile.RangeFamilyAt function and it will now run just the tile list with no check whatsoever. Which means it ignores min/max-zoom parameters and is also hard to kill:
ctrl-c only kills the running seed, not the channel I believe. 

https://github.com/stvno/tegola/blob/seedTileList/cmd/tegola/cmd/cache.go#L312

ARolek:
[...]

Issue #531, opened 2018-10-15
cache command redundantly caches higher zoom tiles when seeding from tile list


ear7h:
Lower zoom tiles will be redundantly cached, for example running tegola cache seed --max-zoom=10 --tile-list=xxx.txt will cache the zero tile (0/0/0) 4 times, once for each time it walks up the tile tree.

xxx.txt:
```
10/495/637
10/496/637
10/495/638
10/496/638
```

PR #533, opened 2018-10-16
Adds the explicit-list option for seeding

stvno:
This PR adds a `--explicit-tile` option that takes a filename with a list of tiles, similar to `--tile-list` it will seed only the tiles in this list. It is based on the discussion in #529

### DIFF
--- a/cmd/tegola/cmd/cache.go
+++ b/cmd/tegola/cmd/cache.go
@@ -29,6 +29,8 @@ var (
 	cacheZXY string
 	// read zxy values from a file
 	cacheFile string
+	// read zxy values from a file for explicit seeding
+	cacheExplicit string
 	//	filter which maps to process. default will operate on all mapps
 	cacheMap string
 	// the min zoom to cache from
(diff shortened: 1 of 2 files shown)


gdey:
The checks are failing because of the redis creds. @ear7h, when you setup redis did we not get creds for it?

ear7h:
@gdey the default config values for the redis cache correspond with the default values in a redis installation. I was also not the one who set up redis on our ci system.

And checking the logs, I think it's the aws cache that is causing the build to fail.

gdey:
@ear7h Yeah, talked to @ARolek , I will have to look at the creds and see what's going on with the ci and Travis.

gdey:
Reading #529 it seems a better approach would be to fix the redundancy issue. Do we have an issue for the redundancy?

ARolek:
@gdey I don't think we need to worry about the parent redundancy issue right now and @ear7h did open an issue around it #531. 

The important thing to fix here is the user experience. If someone calls:

```
$ tegola cache seed --tile-list=tile-list.txt
```
we should only seed their tile list. The inclusion of parent and child tiles should only happen if `--min-zoom` and `--max-zoom` are set. 

We may need to turn `tile-list` into a subcommand of `seed` and `purge` to do this.

PR #535, opened 2018-10-17
Stvno docker

ARolek:
@stvno I found the underlying issue and I also cleaned up a few more parts of the Dockerfile. Check it out when you have a min.


ARolek:
LGTM

PR #536, opened 2018-10-18
Issue 519

tierpod:
This PR adds ability to configure webserver response headers:

```toml
[webserver.headers]
Access-Control-Allow-Origin = "map.example.com"
Cache-Control = "no-cache, no-store, must-revalidate"
```

If "Access-Control-Allow-Origin" and "Access-Control-Allow-Methods" is not set, use default values:
```
Access-Control-Allow-Methods: GET, OPTIONS
Access-Control-Allow-Origin: *
```

Some headers blacklisted and can not be changed (config/config.go: blacklistHeaders)

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,10 @@ Under the `maps` section, map layers are associated with data provider layers an
 [webserver]
 port = ":9090"              # port to bind the web server to. defaults ":8080"
 
+	[webserver.headers]
+	Access-Control-Allow-Origin = "*"
+	Cache-Control = "no-cache, no-store, must-revalidate"
+
 [cache]                     # configure a tile cache
 type = "file"               # a file cache will cache to the local file system
 basepath = "/tmp/tegola"    # where to write the file cache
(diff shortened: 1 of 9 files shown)


ARolek:
@tierpod thanks for the PR! I just reviewed it and overall it looks really good. You hit all the requirements and included tests. Very much appreciated.

I think we can simplify a few aspects of the PR. I made some changes in [another branch](https://github.com/go-spatial/tegola/commit/ec8547cc7bfc299380c1724683b05e563a425675) so you can take a look. Let me know your thoughts.

tierpod:
Yes, your version looks better, thank you! I see that HeadersHandler sets default CORS headers first and then overwrite them if present in configuration. https://github.com/go-spatial/tegola/blob/ec8547cc7bfc299380c1724683b05e563a425675/server/middleware_headers.go#L9-L15

ARolek:
Yeah exactly. Also removes the old CORS middleware so we can keep the header logic in one place. Thanks again for the PR. I'm going to get this merged into v0.8.x

PR #542, opened 2018-10-25
Give units when reporting large tiles. closes #494

### LINKED ISSUE #494: Give units when reporting large tiles
Right now large tiles are reported like `handle_map_layer_zxy.go:177: tile z:8, x:40, y:86 is rather large - 541058` but it doesn't say if that's bytes, kb, or something else.

ARolek:


server/handle_map_layer_zxy.go
```diff
@@ -177,6 +177,6 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// check for tile size warnings
 	if len(pbyte) > MaxTileSize {
-		log.Infof("tile z:%v, x:%v, y:%v is rather large - %v", req.z, req.x, req.y, len(pbyte))
+		log.Infof("tile z:%v, x:%v, y:%v is rather large - %vKb", req.z, req.x, req.y, len(pbyte/1024))
```
gdey: This should be:
```suggestion
		log.Infof("tile z:%v, x:%v, y:%v is rather large - %vKb", req.z, req.x, req.y, len(pbyte)/1024)
```

ARolek: 🤦‍♂️

PR #543, opened 2018-10-25
Improve display of tile rendering times. closes #484

### LINKED ISSUE #484: Improve display of tile rendering times
When seeding, the log shows stuff like

```
2018-08-03 04:38:53 [INFO] cache.go:251: seeding map (bolder) tile (4/15/15) took: 546.098571ms
2018-08-03 04:38:53 [INFO] cache.go:251: seeding map (bolder) tile (4/14/8) took: 4.656699255s
2018-08-03 04:39:06 [INFO] cache.go:251: seeding map (bolder) tile (4/2/5) took: 54.470062566s
```

There are two problems with this

1. The times are shown with a nano-second precision. This is excessive and makes it harder to read. Times to the nearest ms or 0.1 ms are accurate enough

2. The times are in different units, which makes it hard to parse. For many applications I'd like to capture the data and do histograms of tile generation time so I know if a change is an improvement or not. This is easier with consistent units. Even when viewing directly, consistent units makes it easier to see at a glance when a time is larger.

ARolek:


cmd/tegola/cmd/cache.go
```diff
@@ -248,7 +248,7 @@ func SeedWorker(ctx context.Context, mt MapTile) error {
 	//	https://github.com/golang/go/issues/14045 - should be addressed in Go 1.11
 	runtime.GC()
 
-	log.Infof("seeding map (%v) tile (%v/%v/%v) took: %v", mt.MapName, z, x, y, time.Now().Sub(t))
+	log.Infof("seeding map (%v) tile (%v/%v/%v) took: %vms", mt.MapName, z, x, y, time.Now().Sub(t).Nanoseconds()/1000)
```
olt: That's microseconds. Not sure if this precision in necessary :-)
Either Nanoseconds()/1000000 or Seconds()*1000 and %d as format.

ARolek: @olt great catch! 🙏

PR #547, opened 2018-11-01
Changes to explicit tiles list as per #529

gdey:
As per the discussion in #529


gdey:
@stvno would love for you to review this as well. Pretty big refactor.

gdey:
The reason it’s failing on 1.9.x is because `strings.Builder` was introduced in 1.10.x. So, I will need to change this.

ARolek:
@gdey pull request looks good! I made some minor comment, logging and usage instruction formatting changes. All non functional.

PR #556, opened 2018-11-09
Add support for --no-cache override

tierpod:
Add tegola serve [-n, --no-cache] flag. It overrides config cache section.

### DIFF
--- a/cmd/tegola/cmd/root.go
+++ b/cmd/tegola/cmd/root.go
@@ -29,6 +29,7 @@ func init() {
 
 	// server
 	serverCmd.Flags().StringVarP(&serverPort, "port", "p", ":8080", "port to bind tile server to")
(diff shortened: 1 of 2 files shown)


cmd/tegola/cmd/root.go
```diff
@@ -80,7 +81,9 @@ func initConfig(configFile string, cacheRequired bool) (err error) {
 	if len(conf.Cache) == 0 && cacheRequired {
 		return fmt.Errorf("No cache defined in config, please check your config (%v).", configFile)
 	}
-	if len(conf.Cache) > 0 {
+	if serverNoCache {
+		log.Info("Turn off the cache.")
```
ARolek: Let's change this log to "Cache explicitly turned off by user" or something a bit more explicit.

ARolek:
LGTM pending a better log message. Thanks for the clean solution.

tierpod:
No problem. Made the new commit.

PR #560, opened 2018-11-14
Add per-map tile_buffer configuration

tierpod:
* server: replace TileBuffer with TileBuffers,
* config: add per-map tile_buffer configuration
* cmd: use config.TileBuffersMap() for server.TileBuffers

### DIFF
--- a/atlas/map.go
+++ b/atlas/map.go
@@ -9,8 +9,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/golang/protobuf/proto"
-
 	"github.com/go-spatial/geom"
 	"github.com/go-spatial/geom/slippy"
 	"github.com/go-spatial/tegola"
@@ -20,6 +18,7 @@ import (
 	"github.com/go-spatial/tegola/mvt"
 	"github.com/go-spatial/tegola/provider"
 	"github.com/go-spatial/tegola/provider/debug"
+	"github.com/golang/protobuf/proto"
 )
 
 // NewMap creates a new map with the necessary default values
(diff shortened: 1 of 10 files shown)


ARolek:
@tierpod apologies on the slow review. Having configurable tile buffers per map was always intended but never plumbed through. Thank you for taking the time to put this in motion.

I spent some time with this commit today and I think we can accomplish the goal with a slightly different approach. The `atlas` package was positioned to allow for configurable tile buffers for each map. If you take a look at [atlas/map.go](https://github.com/go-spatial/tegola/blob/v0.9.x/atlas/map.go#L26) you will see that a `TileBuffer` property exists on the `Map` struct (this value should probably use the `DefaultTileBuffer` const). The `NewWebMercatorMap()` function does not take in a TileBuffer but the various properties are set during initialization of the map [here](https://github.com/go-spatial/tegola/blob/v0.9.x/cmd/internal/register/maps.go#L59). If the `TileBuffer` value is set on the `Map` during initialization then right before we call [slippy.NewTile()](https://github.com/go-spatial/tegola/blob/v0.9.x/server/handle_map_layer_zxy.go#L141) we can have a piece of logic to determine which `TileBuffer` value to use (i.e. map, global or default). 

Admittedly this approach still has the wart of the `server.TileBuffer` variable, which would probably better fit on the `Atlas` instance but it get's rid of managing a map of buffer values decoupled from their Map instances. 

Let me know what you think.

tierpod:
> apologies on the slow review

No problem, I'm glad to help your project.

> Let me know what you think.

Looks good. I will try to implement it and make new commits in my free time.

tierpod:
@ARolek I'm complete new version. What do you thing about it?

ARolek:
@tierpod Looks great! Thanks for making the adjustments. I'm going to accept the PR and add a few comments to the code and get the CI working again.

PR #564, opened 2018-12-13
Issue 562 - Implement don't clipping geometry

### LINKED ISSUE #562: Option dont_clip for layer
It would be good to have a possibility to disable clipping for a specific layer. Something like _dont_clip = true_ similar to _dont_simplify_.

While this could give speed and size improvements for simple geometries like buildings, main reason is use case described below:

I'm trying to create curved center line labels for water bodies. Algorithm on database side calculates a place where a name of waterbody in largest possible font would fit as well as letter spacing and stuff. This gives my a curved line for each label. Mapbox has already added a line label placement option label-center which should print a label in the center of the given line. But when center line overlaps more than one tile it is split by Tegola, so mapbox-gl-js is trying to put a center label on all parts separately which is not ok, because:
* multiple labels are rendered (when they fit)
* labels of large waterbodies could be too large to fit into one tile (because font size calculated could be 30 or even more)

Example:
image
Here a gray line in the middle of waterbody is a center line. Name "Kauno marios" should be written in the middle of it, but center line is split per tile (marked with red line). Also font size and letter spacing of "Kauno marios" should be larger, but in that case it does not fit in one tile and whole label is not rendered.

With clipping switched off I expect this center line to appear in all involved tiles. Label would be rendered multiple times but in the same spot, so should be ok (I have no way to test it now).

paumas:
Fix #562 issue

### DIFF
--- a/cmd/xyz2svg/cmd/draw.go
+++ b/cmd/xyz2svg/cmd/draw.go
@@ -175,7 +175,7 @@ func drawFeatures(pname string, tiler provider.Tiler, layers []string, gid int, 
 
 			// Clip and validate
 			ext := geom.NewExtent([2]float64{pbb[0], pbb[1]}, [2]float64{pbb[2], pbb[3]})
-			vg, err := validate.CleanGeometry(ctx, sg, ext)
+			vg, err := validate.CleanGeometry(ctx, sg, ext, true)
 
 			// Draw each of the steps.
 			ffname, file, err := dfn.createFile(pname, name, gid, count)
--- a/maths/validate/validate.go
+++ b/maths/validate/validate.go
@@ -96,12 +96,16 @@ func scaleMultiPolygon(p tegola.MultiPolygon, factor float64) (bmp basic.MultiPo
 	return bmp
 }
 
-func CleanGeometry(ctx context.Context, g tegola.Geometry, extent *geom.Extent) (geo tegola.Geometry, err error) {
+func CleanGeometry(ctx context.Context, g tegola.Geometry, extent *geom.Extent, clipGeometry bool) (geo tegola.Geometry, err error) {
 	if g == nil {
 		return nil, nil
 	}
 	switch gg := g.(type) {
 	case tegola.Polygon:
+		if !clipGeometry {
+			extent = nil
+		}
+
 		expp := scalePolygon(gg, 10.0)
 		ext := extent.ScaleBy(10.0)
 		hm := hitmap.NewFromGeometry(expp)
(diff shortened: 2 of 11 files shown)


ARolek:
@gdey ping.

cmd/xyz2svg/cmd/draw.go
```diff
@@ -175,7 +175,7 @@ func drawFeatures(pname string, tiler provider.Tiler, layers []string, gid int,
 
 			// Clip and validate
 			ext := geom.NewExtent([2]float64{pbb[0], pbb[1]}, [2]float64{pbb[2], pbb[3]})
-			vg, err := validate.CleanGeometry(ctx, sg, ext)
+			vg, err := validate.CleanGeometry(ctx, sg, ext, true)
```
ARolek: How about instead of adding in a new parameter we just pass in a nil extent? The Extent methods are designed to check for nil. If we do this, I think we can clean up a bunch of the additions in this PR.

maths/validate/validate.go
```diff
@@ -96,12 +96,16 @@ func scaleMultiPolygon(p tegola.MultiPolygon, factor float64) (bmp basic.MultiPo
 	return bmp
 }
 
-func CleanGeometry(ctx context.Context, g tegola.Geometry, extent *geom.Extent) (geo tegola.Geometry, err error) {
+func CleanGeometry(ctx context.Context, g tegola.Geometry, extent *geom.Extent, clipGeometry bool) (geo tegola.Geometry, err error) {
```
ARolek: remove `clipGeometry` parameter and instead add a comment describing the use of a nil extent to CleanGemetry.

maths/validate/validate.go
```diff
@@ -96,12 +96,16 @@ func scaleMultiPolygon(p tegola.MultiPolygon, factor float64) (bmp basic.MultiPo
...
 	if g == nil {
 		return nil, nil
 	}
 	switch gg := g.(type) {
 	case tegola.Polygon:
+		if !clipGeometry {
```
ARolek: These checks can all go away if we use the nil extent approach.

mvt/feature.go
```diff
@@ -594,7 +594,7 @@ func encodeGeometry(ctx context.Context, geometry tegola.Geometry, tile *tegola.
 	}
 	ext := geom.NewExtent([2]float64{pbb[0], pbb[1]}, [2]float64{pbb[2], pbb[3]})
 
-	geometry, err = validate.CleanGeometry(ctx, sg, ext)
+	geometry, err = validate.CleanGeometry(ctx, sg, ext, clip)
```
ARolek: Use a nil extent rather than plumb the boolean through any further.

ARolek:
This was merged in along with the review comments in #593.

PR #566, opened 2019-01-04
Use a hard-coded zoom level when inspecting geom type

### LINKED ISSUE #343: !ZOOM! substitution error
With a query containing the SQL `way_area > pixel_size(!ZOOM!)*4` I get `ERROR: syntax error at or near "ANY" (SQLSTATE 42601)`

Because this is similar to previous problems with st_asbinary, I checked the code

https://github.com/terranodo/tegola/blob/9c523ad35e1e6ce3a06ca09e580518e78ac65622/provider/postgis/postgis.go#L277-L279

Substituting an expression to where an integer is expected isn't safe. Functions that accept zoom as a parameter are reasonably common, e.g. [ZRes](https://github.com/mapbox/postgis-vt-util/blob/master/src/ZRes.sql), or functions which take something that depends on z, like [LabelGrid](https://github.com/mapbox/postgis-vt-util/blob/master/src/LabelGrid.sql) from [postgis-vt-util](https://github.com/mapbox/postgis-vt-util). Many SQL functions are called with something depending on zoom, e.g. [ST_SnapToGrid](http://postgis.net/docs/manual-2.4/ST_SnapToGrid.html).

Unlike the st_asbinary case, I'm not aware of any workarounds.

russss:
This resolves #343, although not in the most ideal way. I've been
running with this patch on openinframap.org for a few months so I
think it's worth documenting it.

As far as I'm aware, the only situation in which this would cause a
problem is if a layer's SQL has a zoom filter which filters out all
objects at zoom 24, but displays them at lower zooms.

I'm not sure if this is a common situation (it isn't in my
configuration), and I don't necessarily expect this patch to be
merged. I'm providing it so it's available to others as an option, in
the absence of a better solution.

### DIFF
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -431,7 +431,7 @@ func (p Provider) inspectLayerGeomType(l *Layer) error {
 
 	// if a !ZOOM! token exists, all features could be filtered out so we don't have a geometry to inspect it's type.
 	// address this by replacing the !ZOOM! token with an ANY statement which includes all zooms
-	sql = strings.Replace(sql, "!ZOOM!", "ANY('{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24}')", 1)
+	sql = strings.Replace(sql, "!ZOOM!", "24", 1)
 
 	// we need a tile to run our sql through the replacer
 	tile := slippy.NewTile(0, 0, 0, 64, tegola.WebMercator)
--- a/server/README.md
+++ b/server/README.md
@@ -5,13 +5,15 @@ The server package is responsible for handling webserver requests for map tiles 
 ```toml
 [webserver]
 port = ":9090" # set something different than default ":8080"
+
+[webserver.headers]
+Access-Control-Allow-Origin = "*"
 ```
 
 ### Config properties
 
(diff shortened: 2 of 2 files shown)


gdey:
LGTM, but @ARolek should really review this. I will defer to his judgment.

russss:
Additional comment: I reckon the situations where you might want to use `!ZOOM!` as the input to a function, or in other ways where the `ANY()` clause would error, are likely significantly more common than situations where a zoom level of 24 would return no results.  

So I think this change is a net benefit until a better (probably more complex) solution is found.

(Also it looks like the CI is broken here.)

ARolek:
@russss I think this approach would fail for any layer that's not at zoom 24 which is quite common. An alternative work around is to set the `geometry_type` value for the layer which was introduced in v0.7.0. When this is set tegola will no longer try to sniff for the geometry type. You can find documentation [here](https://tegola.io/documentation/configuration/#postgis-1)

russss:
Yes, my maps do tend to show all detail at high zooms but I was curious how common other approaches were. 

I didn't spot the `geometry_type` key, but I think it would still be nice to have this geom type detecton logic work better. At the least, if it fails currently it's not obvious that you can fix it by specifying `geometry_type`.

How about iterating from zoom level 24 down to 1 and using the first row that gets returned?

ARolek:
Closing as this is old and never got merged. Happy to revisit.

PR #567, opened 2019-01-07
make better performance by Remove not needed sort

vahid-sohrabloo:
I think we don't need sort.Ints in FindIntersectsWithEventQueueWithoutIntersect function.
It makes decrease 30%-50%  time to render tiles for me.


ARolek:
@vahid-sohrabloo great find. I talked with @gdey about this and I believe the sorting is used for tests. That considered we should move sorting to the test, especially with the performance improvement you're reporting.

gdey:
LGTM

Issue #573, opened 2019-02-17
tile_buffer not configured correctly with multiple maps


henrythasler:
branch: v0.9.x
When configuring a map-specific tile_buffer, the setting is not applied to every map in the for-loop as it returns on first hit.

This diff solved this particular issue:

```
diff --git a/config/config.go b/config/config.go
index c405a62..32d22b6 100644
--- a/config/config.go
+++ b/config/config.go
@@ -147,13 +147,13 @@ func (c *Config) ConfigureTileBuffers() {
                // if there is a tile buffer config for this map, use it
                if m.TileBuffer != nil {
                        c.Maps[mapKey].TileBuffer = m.TileBuffer
-                       return
+                       continue
                }
 
                // if there is a global tile buffer config, use it
                if c.TileBuffer != nil {
                        c.Maps[mapKey].TileBuffer = c.TileBuffer
-                       return
+                       continue
                }
 
                // tile buffer is not configured, use default
```

ARolek:
@henrythasler thanks for the patch. I have applied it and updated the test cases. Merged into v0.9.x

PR #577, opened 2019-02-26
Fixed tile buffer not being set correctly #573

ARolek:
Addressed issue where map tile buffers were not being set correctly
for configured maps 2+. This was due to a return statement when
a continue should have been used. Updated test cases accordingly.


gdey:
LGTM

PR #592, opened 2019-04-05
Issue 429 handle empty geometry collection

ARolek:


gdey:
I don't think this is the right place for this check. Why is it only the PostGIS provider is doing this check. I think it should be higher up in the call stack. The provider should just return an empty collection. The thing that gets the features from the provider should then decide what to do with the collection if it's empty or not.

ARolek:
@gdey I moved the check to atlas/map.go and added in a testcase.

PR #595, opened 2019-04-09
Update sqlite

gdey:
This updates the sqlite dependence.


provider/gpkg/gpkg_register_internal_test.go
```diff
@@ -1107,8 +1109,6 @@ func TestCleanup(t *testing.T) {
 
 	for name, tc := range tests {
 		tc := tc
```
ARolek: leftover

PR #596, opened 2019-04-09
Custom webserver headers are not added to OPTIONS requests

### LINKED ISSUE #594: User defined http response headers are not added to OPTIONS requests
When custom webserver response headers are configured they're added to the all requests except `OPTIONS`

ARolek:
* Fixed bug where user defined headers were not being added
to OPTIONS requests.
* Added associated tests.

closes #594

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -151,5 +151,8 @@ var URLRoot = func(r *http.Request) string {
 func corsHandler(w http.ResponseWriter, r *http.Request, params map[string]string) {
 	w.Header().Set("Access-Control-Allow-Origin", CORSAllowedOrigin)
 	w.Header().Set("Access-Control-Allow-Methods", CORSAllowedMethods)
+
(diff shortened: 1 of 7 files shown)


ARolek:
@gdey I refactored this a bit per your suggestions.

server/server.go
```diff
@@ -154,18 +155,20 @@ func corsHandler(w http.ResponseWriter, r *http.Request, params map[string]strin
...
 	}
 
 	// set user defined headers
-	for name, value := range Headers {
-		v, ok := value.(string)
-		if ok {
-			w.Header().Set(name, v)
+	for name, val := range Headers {
+		if val == "" {
+			log.Warn("header (%v) has no value")
```
gdey: you are missing the val, value in the Warn here.

It should be
```go
log.Warnf("header (%v) had no value",val)
```

ARolek: Good catch.

Issue #600, opened 2019-04-12
cache/redis: add configurable expiration time


tierpod:
I see that RedisCache has Expiration field: https://github.com/go-spatial/tegola/blob/master/cache/redis/redis.go#L90

It would be nice to add ability to configure expiration time. I will try to implement it.

```go
func (c *Client) Set(key string, value interface{}, expiration time.Duration) *StatusCmd
```

ARolek:
This has been merged into the v0.10.x branch

PR #601, opened 2019-04-12
cache/redis: add configurable key expiration time #600

tierpod:
PR for issue #600


ARolek:
LGTM. Thank!

ARolek:
@tierpod I have not merged this in yet because I'm debating internally if the config value should be `expires` or `ttl`. I'm starting to lean towards `ttl` for "Time To Live" as I think that better describes the value the user is setting. To me, `expires` seems like a specific time that the cache should expire the entry, rather than a time after setting the entry in the cache.

Thoughts?

tierpod:
I agree with you. At the beginning I made "ttl" value but then I was a little confused with "Expiration" argument in redis.Set function. I have changed it today.

ARolek:
@tierpod great! Thanks for the change.

PR #608, opened 2019-06-19
Upgraded internal viewer Mapbox GL JS to 1.0.0

ARolek:


server/static/js/tegola.js
```diff
@@ -69,10 +69,24 @@ var app = new Vue({
...
+							color = this.map.getPaintProperty(layers[j].name, 'line-color');
+						} catch(e){}
+
+						try {
+							color = this.map.getPaintProperty(layers[j].name, 'fill-outline-color');
+						} catch(e){}
+
+						try {
+							color = this.map.getPaintProperty(layers[j].name, 'circle-color');
+						} catch(e){}
```
gdey: What is removed is not logically the same as what got added. In the added code the color variable will attempt to be set all three times. Doesn't that make it so that only the last getPoaintPropterty is applied?

gdey: Another possible solution:
```js
var properties = ['line-color','fill-outline-color','circle-color']
properties.Some(func(p){
      try{
          color = this.map.getPaintProperty(layers[j].name, p);
          return true;
     } catch(e){}
     return false;
})
```

ARolek: Thanks for the snippet. I pushed up something similar.

PR #611, opened 2019-07-05
Content-Length for non-gzipped responses

thomersch:
- Corrected Content-Length
- also skipping the unnecessary buffer


ARolek:
@thomersch this looks great, thank you! Can you adjust the PR so it's against the `v0.10.x` branch rather than `master`?

PR #612, opened 2019-07-10
[ci] Currently Travis breaks because of S3

gdey:
Travis breaks tests on pull requests because it is unable to decrypt
the creds for S3. We should really skip these test, and not fail.


ARolek:
LGTM

PR #614, opened 2019-07-22
Auto config for PostGIS, pt 1

stdmn:
Get tables from Postgres information_schema and automatically create layers. Allows for layers grouped to a map using regex.

High level changes:
- Under providers, add new "auto" option
    - Option to create an auto_config for each provider which will add specified options to automatically-generated provider layers
    - Auto config is optional. Defaults will be used if not provided
- Maps created using regex in basic format "Provider.<regex>" for auto providers
    - Map layers automatically created for auto layers using tables that match regex

### DIFF
--- a/cmd/internal/register/maps.go
+++ b/cmd/internal/register/maps.go
@@ -105,7 +105,6 @@ func Maps(a *atlas.Atlas, conf config.Config, providers map[string]provider.Tile
 
 			// lookup our proivder
 			provider, ok := providers[providerLayer[0]]
-			fmt.Println("printing provider:", provider)
 
 			// determine if layers should automatically be created from provider regex
 			auto := false
--- a/cmd/internal/register/maps.go
+++ b/cmd/internal/register/maps.go
@@ -3,6 +3,8 @@ package register
 import (
 	"fmt"
 	"html"
+	"log"
+	"regexp"
 	"strings"
 
 	"github.com/go-spatial/geom"
@@ -56,7 +58,11 @@ func (e ErrDefaultTagsInvalid) Error() string {
 }
 
 // Maps registers maps with with atlas
-func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler) error {
+// note that we are pulling the full config file to get both providers (to check if auto) and maps
+
+func Maps(a *atlas.Atlas, conf config.Config, providers map[string]provider.Tiler) error {
+
+	maps := conf.Maps
 
 	// iterate our maps
 	for _, m := range maps {
@@ -99,12 +105,24 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
 
 			// lookup our proivder
 			provider, ok := providers[providerLayer[0]]
+			fmt.Println("printing provider:", provider)
+
+			// determine if layers should automatically be created from provider regex
+			auto := false
+			for _, p := range conf.Providers {
+				if p["name"] == providerLayer[0] && p["auto"] == true {
+					auto = true
+				}
+			}
+
 			if !ok {
 				return ErrProviderNotFound{providerLayer[0]}
 			}
 
 			// read the provider's layer names
 			layerInfos, err := provider.Layers()
+			fmt.Println("printing provider layers:", layerInfos)
+
 			if err != nil {
 				return ErrFetchingLayerInfo{
 					Provider: providerLayer[0],
@@ -113,16 +131,47 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
 
 			// confirm our providerLayer name is registered
 			var found bool
+			// this must be an array because auto providers may have multiple layers that match regex
+			var provLayers []string
 			var layerGeomType tegola.Geometry
-			for i := range layerInfos {
-				if layerInfos[i].Name() == providerLayer[1] {
-					found = true
+			for i, info := range layerInfos {
+				// check if provider layers are automatically generated
+				if auto {
+					// regex to check against. ^ to match only beginning of phrase
+					exp := fmt.Sprintf("^%v", providerLayer[1])
+					// if * (wildcard), select all
+					if exp == "^*" {
+						provLayers = append(provLayers, info.Name())
+						found = true
+						layerGeomType = info.GeomType()
+					} else {
+						// must compile the regex first before testing phrase
+						r, err := regexp.Compile(exp)
+						if err != nil {
+							log.Printf("Error when parsing regex (layer: %v): %v", info.Name(), err)
+						} else {
+							// if regex matches, push provider layer. Note that there is no break--we need to find all layers that match
+							if r.MatchString(info.Name()) {
+								provLayers = append(provLayers, info.Name())
+								found = true
+								// because this is the same for all auto provider layers, we don't need to worry about which loop we are on
+								layerGeomType = info.GeomType()
+							}
+						}
+					}
 
-					// read the layerGeomType
-					layerGeomType = layerInfos[i].GeomType()
-					break
+				} else {
+					if layerInfos[i].Name() == providerLayer[1] {
+						found = true
+						provLayers = append(provLayers, providerLayer[1])
+						// read the layerGeomType
+						layerGeomType = layerInfos[i].GeomType()
+						break
+
+					}
 				}
 			}
+
 			if !found {
 				return ErrProviderLayerNotRegistered{
 					MapName:       string(m.Name),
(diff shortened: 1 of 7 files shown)


provider/postgis/postgis.go
```diff
@@ -199,14 +221,49 @@ func NewTileProvider(config dict.Dicter) (provider.Tiler, error) {
...
+
+		if err != nil {
+			return nil, err
+		}
+		// generate layers from information_schema of provider
+		autoTables, err = genLayers(p.pool)
+		for i := 0; i < len(autoTables); i++ {
+
+			// this is very hacky. I was having problems creating an empty dict.Dicter
+			defaultLayer, err := config.Map("")
```
stdmn: Kept getting errors when I tried to make this an empty dict.Dicter. This is a hack to try to fix that

provider/postgis/postgis.go
```diff
@@ -199,14 +221,49 @@ func NewTileProvider(config dict.Dicter) (provider.Tiler, error) {
...
+			if err != nil {
+				log.Println(err)
+			}
+			if len(autoConfig) > 0 {
+				// use auto config details to fill in layer info
+				defaultLayer = autoConfig[0]
+			}
+			// add auto config layer
```
stdmn: Ideally, we could add the layer name to this dict.Dicter element, but I wasn't sure how to do that once the mapSlice has been created. This affects line 261 below

provider/postgis/postgis.go
```diff
@@ -199,14 +221,49 @@ func NewTileProvider(config dict.Dicter) (provider.Tiler, error) {
...
 	for i, layer := range layers {
 
-		lname, err := layer.String(ConfigKeyLayerName, nil)
-		if err != nil {
-			return nil, fmt.Errorf("For layer (%v) we got the following error trying to get the layer's name field: %v", i, err)
+		var lname string
+
+		// if auto, get layername from automatically generated layernames
+		if auto {
+			lname = autoTables[i]
```
stdmn: See previous comment

cmd/internal/register/maps.go
```diff
@@ -99,12 +105,24 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
...
+
+			// determine if layers should automatically be created from provider regex
+			auto := false
+			for _, p := range conf.Providers {
+				if p["name"] == providerLayer[0] && p["auto"] == true {
+					auto = true
+				}
+			}
+
 			if !ok {
```
gdey: This should be right after 107. Before the fmt.Println

cmd/internal/register/maps.go
```diff
@@ -99,12 +105,24 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
 
 			// lookup our proivder
 			provider, ok := providers[providerLayer[0]]
+			fmt.Println("printing provider:", provider)
+
+			// determine if layers should automatically be created from provider regex
+			auto := false
+			for _, p := range conf.Providers {
+				if p["name"] == providerLayer[0] && p["auto"] == true {
+					auto = true
```
gdey: I thought we got rid of the `auto`. I don't think we need it. I think this is unnecessary. The maps sections should not care if the Provider is auto-configured or not. It should just care about the layers it needs and it can do this by looking at the layer name to see if it returns multi-layers.

cmd/internal/register/maps.go
```diff
@@ -99,12 +105,24 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
```
gdey: I would have expected the above at line 96 to have been changed to a `strings.SplitN(string(l.ProviderLayer),".",2)` (https://godoc.org/strings#SplitN) , this would be to make sure the name of the layer does not get messed up.

cmd/internal/register/maps.go
```diff
@@ -113,16 +131,47 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
...
 
-					// read the layerGeomType
-					layerGeomType = layerInfos[i].GeomType()
-					break
+					}
 				}
 			}
+
 			if !found {
 				return ErrProviderLayerNotRegistered{
```
gdey: We should be able to get rid of the `found` variable and just check to see if `len(provLayers) > 0`.

cmd/internal/register/maps.go
```diff
@@ -113,16 +131,47 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
...
 			var found bool
+			// this must be an array because auto providers may have multiple layers that match regex
+			var provLayers []string
 			var layerGeomType tegola.Geometry
-			for i := range layerInfos {
-				if layerInfos[i].Name() == providerLayer[1] {
-					found = true
+			for i, info := range layerInfos {
+				// check if provider layers are automatically generated
+				if auto {
```
gdey: Don't know why we need this check.The expression: `^san_deigo` would still match `san_diego` as the name. So, why not just have two cases here:
```go

if providerLayer[1] == "*" {
  // return all the layers:
    provLayers = append(provLayers,info.Name())
    found = true // Do we need this now as len(provLayers) is the same?
    layerGeoType = info.GeoType()
    continue
} 
r, err := regex.Compile("^"+providerLayer[1])
if err != nil {
       log.Printf("Error when parsing regex (layer: %v): %v", info.Name(), err)
       continue Provider // add a Providers label at 111
}
if !r.MatchString(info.Name()) {
    continue
}
provLayers = append(provLayers, info.Name())
found = true
layerGeoType = info.GeoType()

```

cmd/internal/register/maps.go
```diff
@@ -152,18 +201,29 @@ func Maps(a *atlas.Atlas, maps []config.Map, providers map[string]provider.Tiler
...
+			// for each provider layer, get map name
+			// this is a loop to capture auto provider layers with multiple layers that match regex
+			for _, name := range provLayers {
+				var lname string
+				if auto {
+					lname = name
+				} else {
+					lname = string(l.Name)
+				}
+
```
gdey: Why use `l.Name` and not just the name, value.

cmd/tegola/cmd/root.go
```diff
@@ -75,7 +78,8 @@ func initConfig(configFile string, cacheRequired bool) (err error) {
 	}
 
 	// init our maps
-	if err = register.Maps(nil, conf.Maps, providers); err != nil {
+	// note that we are sending the whole config file to include both maps and providers
+	if err = register.Maps(nil, conf, providers); err != nil {
```
gdey: I don't think you need to send the whole map in. I think we can still send in just `conf.Maps`

cmd/tegola_lambda/main.go
```diff
@@ -59,7 +59,8 @@ func main() {
 	}
 
 	// register the maps
-	if err = register.Maps(nil, conf.Maps, providers); err != nil {
+	// note that we are sending the whole config file to include both maps and providers
+	if err = register.Maps(nil, conf, providers); err != nil {
```
gdey: Same as before

provider/postgis/layer.go
```diff
@@ -16,6 +16,8 @@ type Layer struct {
 	geomType geom.Geometry
 	// The SRID that the data in the table is stored in. This will default to WebMercator
 	srid uint64
+	// Whether layer tables are automatically configured based on postgres information_schema
+	auto bool
```
gdey: I don't think we need this.

provider/postgis/postgis.go
```diff
@@ -128,6 +143,12 @@ func NewTileProvider(config dict.Dicter) (provider.Tiler, error) {
 		return nil, err
 	}
 
+	auto := DefaultAuto
+	auto, err = config.Bool(ConfigKeyAuto, &auto)
+	if err != nil {
+		return nil, err
+	}
+
```
gdey: Can't we just check to see it there are any layers defined? And if there aren't then assume auto-config.

cmd/internal/register/maps.go
stdmn: @ARolek Mentioned that he wanted it to automatically configure even if there is no auto_config present. This requires an auto tag. We can't have both

cmd/tegola/cmd/root.go
stdmn: You'll have to help me with this. I couldn't figure out another way to access the "auto" or "auto_config" on the conf.Providers.

cmd/internal/register/maps.go
stdmn: Can you explain this: `continue Provider // add a Providers label at 111`

I'm a golang newbie (mostly use JS) so not sure what you mean by Providers label etc

cmd/internal/register/maps.go
stdmn: Because there are different conditions based on whether it is auto-generated or not. For a defined layer, Name and ProviderLayerName will be different. But for an auto-generated name, they should be the same (because otherwise the ProviderLayerName would be in regex).

provider/postgis/postgis.go
gdey:
```go
var dict dict.Dict

or 
dict.Dict{}

```

cmd/tegola/cmd/root.go
ARolek: What if conf.Maps was generated and then passed in?

stdmn: Yea. That may be necessary. I still haven't sorted through the whole "No Maps Defined" scenario yet

gdey:
Should we close this as your #615 covers the same ground?

ARolek:
@gdey I think so. no need to keep open 2 PRs. Let's use the one and consolidate the idea there.

Issue #618, opened 2019-07-26
Upgrade geom package to latest version


ear7h:
There are some breaking changes in the `geom:master` and this needs to be taken care of in the latest migration even if it means adding a level of complexity to certain parts of the code base.

The breaking change for this, revolves around the `slippy.Tile` which no longer satisfies the `provider.Tile` interface. The best[1] solution is to create an `atlas.Tile` which would satisfy the `provider.Tile` interface. The ideal solution would be solving #499. 

[1] this is the best solution given our current trajectory of cleaning the rendering bugs with a new simplification algorithm.

ear7h:
update: the `atlas.Tile` cannot be used because the `atlas` package would have to be imported  in `provider` creating a circular dependency. A `provider.tile_t` was created instead.

See the PR referenced below...

PR #619, opened 2019-07-26
Update geom package to latest version

ear7h:
See #618 for more details, but this PR updates the vendored package to `geom:master`. I also added some comments to the relevant data structures to remind other programmers of the direction in which this code base is heading. The roadmap of this repo seems to be the guiding factor of how and when to make changes, so I felt it was important to start documenting the intent (however temporary) of those data structures. Let me know if they should be removed or re-worded.

### DIFF
--- a/atlas/tile.go
+++ b/atlas/tile.go
@@ -1,2 +1,0 @@
-package atlas
-
--- a/atlas/tile.go
+++ b/atlas/tile.go
@@ -1,7 +1,2 @@
 package atlas
 
-type Tile struct {
-	Z uint64
-	X uint64
-	Y uint64
-}
--- a/cmd/tegola/cmd/cache/worker.go
+++ b/cmd/tegola/cmd/cache/worker.go
@@ -27,7 +27,7 @@ func (s seedPurgeWorkerTileError) Error() string {
 	return fmt.Sprintf("error %v tile (%+v): %v", cmd, s.Tile, s.Err)
 }
 
-func seedWorker(tileBuffer *float64, overwrite bool) func(ctx context.Context, mt MapTile) error {
+func seedWorker(overwrite bool) func(ctx context.Context, mt MapTile) error {
 	return func(ctx context.Context, mt MapTile) error {
 		// track how long the tile generation is taking
 		t := time.Now()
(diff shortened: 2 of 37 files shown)


atlas/tile.go
```diff
@@ -1,7 +1,2 @@
 package atlas
```
ARolek: let's delete this file if it's empty

cmd/tegola/cmd/cache/worker.go
```diff
@@ -74,11 +74,6 @@ func seedWorker(tileBuffer *float64, overwrite bool) func(ctx context.Context, m
 			}
 		}
 
-		//	set tile buffer if it was configured by the user
-		if tileBuffer != nil {
```
ARolek: This edit takes away functionality where the buffer can be set at the root of the config. We may deprecate this functionality in the future but we should continue supporting it for now to avoid a regression. `seedWorker` accepts a `tileBuffer`. You could add the following as a replacement:

```go
// check if a global tileBuffer has been set, and if so only set on the map if the default buffer is configured
if tileBuffer != nil && m.TileBuffer != tegola.DefaultTileBuffer {
  m.TileBuffer = tileBuffer
} 
```

ear7h: Thanks for pointing this out, I was mindlessly refactoring... It might be a bigger issue because:
`tileBuffer` is being set by the value in a configuration file (in the caller):
https://github.com/go-spatial/tegola/blob/be45ebde2dae7413f0970a631faffd8e6986c4ca/cmd/tegola/cmd/cache/seed_purge.go#L131-L135
and the map handles it's own tile buffer, as of #560 so, I'm thinking the `tileBuffer` parameter for this function is now unnecessary.

ARolek: The tile buffer can be set at the [config root](https://github.com/go-spatial/tegola/blob/master/config/config.go#L25) as well as the map. The patch I suggested maintains support for the config root approach, which we may eventually deprecate but not right now.

ARolek: I stand corrected, this is handled during config load: https://github.com/go-spatial/tegola/blob/master/config/config.go#L147

ear7h: After some extra-githubular discussions, it was decided to clean up `seedWorker` as part of this pull request as well

Issue #622, opened 2019-07-27
remove tegola.Geometry type from basic.ToWebMercator


ear7h:
the current signature for `basic.ToWebMercator` is:
https://github.com/go-spatial/tegola/blob/c6139463db3230ffe75ae2b3b8b3801167f1b6c7/basic/geometry_math.go#L153
it should be:
```go
func ToWebMercator(SRID uint64, geometry geom.Geometry) (G, error)
```

PR #625, opened 2019-07-31
Remove tegola.Geometry from use in basic.XXXWebMercator functions

ear7h:
see #622 

The original intent was to only change `basic.ToWebMercator` but this change propagated to cloning and walking functions within the basic package. That change propagated to `basic.FromWebMercator`, which I initially didn't want to change (see the first commit). But only fixing one of these functions seemed ugly and the fix for both didn't take much involvement.


basic/geometry_math.go
```diff
@@ -6,151 +6,202 @@ import (
...
 )
 
 // ApplyToPoints applys the given function to each point in the geometry and any sub geometries, return a new transformed geometry.
-func ApplyToPoints(geometry tegola.Geometry, f func(coords ...float64) ([]float64, error)) (G, error) {
+func ApplyToPoints(geometry geom.Geometry, f func(coords ...float64) ([]float64, error)) (geom.Geometry, error) {
 	switch geo := geometry.(type) {
 	default:
-		return G{}, fmt.Errorf("Unknown Geometry: %+v", geometry)
-	case tegola.Point:
+		return nil, fmt.Errorf("Unknown Geometry: %+v", geometry)
```
ARolek: Should this to `%t` so we print the type? Also, let's lowercase the first letter to match the format of the other comments.

basic/geometry_math.go
```diff
@@ -6,151 +6,202 @@ import (
...
-				return G{}, fmt.Errorf("Got error converting line(%v) of multiline: %v", i, err)
+				return nil, fmt.Errorf("got error converting line(%v) of multiline: %v", i, err)
 			}
-			if !geoLine.IsLine() {
-				panic("We did not get the conversion we were expecting")
+
+			// get the value
+			linev, ok := linei.(geom.LineString)
+			if !ok {
+				panic("we did not get the conversion we were expecting")
```
ARolek: @gdey do you think we should keep the panic here or return an error?

basic/geometry_math.go
```diff
@@ -6,151 +6,202 @@ import (
...
-		return G{}, fmt.Errorf("Unknown Geometry: %+v", geometry)
-	case tegola.Point:
-		return G{Point{geo.X(), geo.Y()}}, nil
-	case tegola.Point3:
-		return G{Point3{geo.X(), geo.Y(), geo.Z()}}, nil
-	case tegola.MultiPoint:
-		var pts MultiPoint
-		for _, pt := range geo.Points() {
-			pts = append(pts, Point{pt.X(), pt.Y()})
+		return nil, fmt.Errorf("Unknown Geometry: %+v", geometry)
```
ARolek: maybe use `%t` to print the type. lowercase log text.

basic/geometry_math.go
```diff
@@ -6,151 +6,202 @@ import (
...
 		}
-		return G{pts}, nil
-	case tegola.LineString:
-		var line Line
-		for _, ptGeo := range geo.Subpoints() {
-			line = append(line, Point{ptGeo.X(), ptGeo.Y()})
+		return pts, nil
+
+	case geom.LineString:
+		line := make(geom.LineString, len(geo))
```
ARolek: great optimization over the `append` strategy.

mvt/feature.go
```diff
@@ -519,11 +519,16 @@ func createDebugFile(min, max maths.Pt, geo tegola.Geometry, err error) {
 		return
 	}
 	filename := fmt.Sprintf("/tmp/testcase_%v_%p.json", fln, geo)
-	bgeo, err := basic.CloneGeometry(geo)
+	geo, err = basic.CloneGeometry(geo)
 	if err != nil {
 		log.Errorf("failed to clone geo for test case. %v", err)
 		return
 	}
+	bgeo, err := convert.ToTegola(geo)
```
ARolek: This goes away in https://github.com/go-spatial/tegola/pull/624 so we will have a conflict to deal with. should be minor.

basic/geometry_math.go
ear7h: Using `%T` is a good idea, I'll change that and the casing.

basic/geometry_math.go
gdey: If it gets to this point something with the code is wrong. And nothing we can do is recoverable. So, I think it should panic.

basic/geometry_math.go
```diff
@@ -6,151 +6,202 @@ import (
...
-		for i, line := range geo.Sublines() {
-			geoLine, err := ApplyToPoints(line, f)
+		return lines, nil
+
+	case geom.Polygon:
+		poly := make(geom.Polygon, len(geo))
+
+		for i, line := range geo {
+			// getting a geometry inteface back
+			linei, err := ApplyToPoints(geom.LineString(line), f)
```
gdey: minor typo: `inteface` should be `interface`

Issue #628, opened 2019-08-01
remove tegola.Geometry type from Map.Encode


ear7h:
The following functions have to be converted:
- [x] `basic.ToWebMercator` issue: #622, pr: #625 
- [x] `simplify.SimplifyGeometry`  pr: #648
- [x] `mvt.ScaleGeo` pr: #629
- [x] `validate.CleanGeometry` \* commit: 4136aed8f86c8fbc268c7e757ffb0ba37ccd1376

~Afterwards, the `mvt` package as a whole should be cleaned.~ #629 

\* clean geometry, as it stands, will probably not be converted. Rather, the new `makevalid` from https://github.com/go-spatial/geom will be used. For the purposes of this issue, we will convert around the `validate.CleanGeometry`.

ARolek:
@ear7h I think you should skip validate.CleanGeometry as that what's @gdey is working on. I don't think we benefit from that porting effort. Instead, convert from and to the geom types around that step.

ear7h:
@ARolek I was just in the proccess of updating my original comment when you sent this, I included it in the list for information.

PR #629, opened 2019-08-02
remove basic types from mvt package

ear7h:
See more on #628 

There is a file, `mvt/debug.go` that still uses `basic` types, I'm not sure if it should be kept; please advise.

### DIFF
--- a/mvt/feature.go
+++ b/mvt/feature.go
@@ -4,9 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-spatial/geom"
 	"github.com/go-spatial/geom/encoding/wkt"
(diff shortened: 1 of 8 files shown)


mvt/feature.go
```diff
@@ -4,9 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-spatial/geom"
 	"github.com/go-spatial/geom/encoding/wkt"
-	"github.com/go-spatial/tegola"
-	"github.com/go-spatial/tegola/internal/convert"
 	"github.com/go-spatial/tegola/internal/log"
```
ARolek: let's remove all instances of the tegola internal logger so we can position this for moving to the geom package.

PR #632, opened 2019-08-13
mvt: remove local mvt package and vendor from geom's

ear7h:
pretty straight forward:
* `rm -rf mvt`
* `govendor fetch github.com/geom/...`
* `govendor fetch github.com/geom/encoding/mvt/...`
* `sed -i '' 's/\/tegola\/mvt/\/geom\/encoding\/mvt/g' $(grep 'tegola/mvt' -Irl .)`
* change the `ScaleGeo` line in `atlas/map.go` to `PrepareGeo`
* `goimports -w $(git status -s | grep -v vendor | awk '{ print $2 }')` for styling

~Should I also remove `tegola.Tile` from this repository? only the following files use it:~ I've realized that the methods would need to be replaced too and that is more involved than this pr.
```text
./cmd/tegola/cmd/cache/worker.go
./atlas/map.go
./atlas/atlas.go
./tile_test.go
```


ARolek:
@ear7h can you open up another issue around purging the tegola.Tile from the codebase per your above notes?

Issue #633, opened 2019-08-20
server: checking if map contains tile should be based on intersection not containment


ear7h:
In `master` if a tile does not lie _entirely_ inside the map, then the server returns an error. I don't think this is desired behavior. If any part of a map lies within the tile, I think it should be returned. 

I propose that we change this if statement:
https://github.com/go-spatial/tegola/blob/c6139463db3230ffe75ae2b3b8b3801167f1b6c7/server/handle_map_layer_zxy.go#L144-L148
to:
```go
if _, intersect := m.Bounds.Intersect(textent); !intersect {
```

There is a slight problem with this: that extra computations and allocations take place. A `DoesIntersect` function might be nice.

PR #634, opened 2019-08-21
server: return tiles that lie on a Map's boundary

ear7h:
See #633 
~todo: add tests~


ear7h:
@ARolek  can this be merged? Should I squash and force push?

PR #640, opened 2019-08-28
Added configurable webserver URI Prefix to

### LINKED ISSUE #136: Allow path prefix to be configured
When running behind a reverse proxy, endpoints may be prefixed (e.g. `/tiles/`). Tegola should have the ability to provide an alternate path prefix to match the proxied prefix when it doesn't match Tegola's default (`/`).

(As both an option in the config TOML and as a command line arg.)

ARolek:
Added the webserver config property uri_prefix which allows for
setting an API route prefix. This is commonly used when tegola
is run behind a reverse proxy (i.e. example.com/tegola). Certain
routes (i.e. capabilities) build URLs which are included in the
response and then passed back to the proxy. Without the ability
to set a URI prefix, the URLs in the capabilities responses
will not be useable by the consuming client.

Setting the webserver hostname now ignores the server port. It's
assumed that if the user is setting the hostname they want full
control of it.

closes #136

### DIFF
--- a/server/handle_capabilities_test.go
+++ b/server/handle_capabilities_test.go
@@ -246,61 +246,6 @@ func TestHandleCapabilities(t *testing.T) {
 				},
 			},
 		},
-		"config set hostname unset port": {
-			// With hostname set and port unset in config, urls should have host from config and
-			//  port from uri: "cdn.tegola.io:8080"
-			hostname: "cdn.tegola.io",
-			uri:      "http://localhost:8080/capabilities?debug=true",
-			expected: server.Capabilities{
-				Version: serverVersion,
-				Maps: []server.CapabilitiesMap{
-					{
-						Name:         "test-map",
-						Attribution:  "test attribution",
-						Center:       [3]float64{1.0, 2.0, 3.0},
-						Bounds:       tegola.WGS84Bounds,
-						Capabilities: "http://cdn.tegola.io:8080/capabilities/test-map.json?debug=true",
-						Tiles: []string{
-							"http://cdn.tegola.io:8080/maps/test-map/{z}/{x}/{y}.pbf?debug=true",
-						},
-						Layers: []server.CapabilitiesLayer{
-							{
-								Name: testLayer1.MVTName(),
-								Tiles: []string{
-									fmt.Sprintf("http://cdn.tegola.io:8080/maps/test-map/%v/{z}/{x}/{y}.pbf?debug=true", testLayer1.MVTName()),
-								},
-								MinZoom: testLayer1.MinZoom,
-								MaxZoom: testLayer3.MaxZoom, // layer 1 and layer 3 share a name in our test so the zoom range includes the entire zoom range
-							},
-							{
-								Name: "test-layer-2-name",
-								Tiles: []string{
-									fmt.Sprintf("http://cdn.tegola.io:8080/maps/test-map/%v/{z}/{x}/{y}.pbf?debug=true", testLayer2.MVTName()),
-								},
-								MinZoom: testLayer2.MinZoom,
-								MaxZoom: testLayer2.MaxZoom,
-							},
-							{
-								Name: "debug-tile-outline",
-								Tiles: []string{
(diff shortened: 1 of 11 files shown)


ear7h:
Could you elaborate on the following?
> Setting the webserver hostname now ignores the server port. It's
> assumed that if the user is setting the hostname they want full
> control of it.

ARolek:
@ear7h previously if the hostname was set and a port was included in the request, the the port was appended to the hostname. For example, if I set the hostname to `tegola.io` and a request came in on port :8080 then the host would be modified to `tegola.io:8080`. This is problematic when tegola sits behind a reverse proxy as tegola will likely be listening on an internal port and have no idea what port the request originated on. The user configuring the service would likely know.

server/handle_capabilities_test.go
```diff
@@ -246,61 +246,6 @@ func TestHandleCapabilities(t *testing.T) {
...
-			uri:      "http://localhost:8080/capabilities?debug=true",
-			expected: server.Capabilities{
-				Version: serverVersion,
-				Maps: []server.CapabilitiesMap{
-					{
-						Name:         "test-map",
-						Attribution:  "test attribution",
-						Center:       [3]float64{1.0, 2.0, 3.0},
-						Bounds:       tegola.WGS84Bounds,
-						Capabilities: "http://cdn.tegola.io:8080/capabilities/test-map.json?debug=true",
```
ear7h: If I understand this correctly, shoud this test case stay, but remove the `:8080` from urls?

server/server_internal_test.go
```diff
@@ -52,7 +55,7 @@ func TestHostName(t *testing.T) {
 			// Hostname set, no port in config, but port in url.  Expect <config_host>:<url_port>.
```
ear7h: The `:<url_port>` part should be removed

server/server_internal_test.go
```diff
@@ -52,7 +55,7 @@ func TestHostName(t *testing.T) {
...
 			url:      "http://localhost:8080/capabilities",
 			hostName: "cdn.tegola.io",
-			expected: "cdn.tegola.io:8080",
+			expected: "cdn.tegola.io",
 		},
```
ear7h: I think there should be a test for setting the port and hostname. It would resolve in just the hostname, right?

server/server_test.go
```diff
@@ -15,7 +18,7 @@ import (
 // test server config
 const (
 	httpPort       = ":8080"
-	serverVersion  = "0.4.0"
+	serverVersion  = "0.10.0"
```
ear7h: Wow, nice catch

server/server_test.go
```diff
@@ -90,6 +93,45 @@ func doRequest(a *atlas.Atlas, method string, uri string, body io.Reader) (w *ht
 	return w, router, nil
 }
 
+func TestURLRoot(t *testing.T) {
```
ear7h: Can you move this function below `init()`?

server/server.go
```diff
@@ -142,8 +124,32 @@ func scheme(r *http.Request) string {
...
+	if q != "" {
+		// prepend our query identifier
+		q = "?" + q
+	}
+
+	// usually the url.URL package would be used for building the URL, but the
+	// uri template for the tiles contains characters that the package does not like:
+	// {z}/{x}/{y}. These values are escaped during the String() call which does not
+	// work for the capabilities URLs.
+	return fmt.Sprintf("%v%v%v", URLRoot(r), path.Join(URIPrefix, uri), q)
```
ear7h: here you can just do `path.Join(URIPrefix, uriParts...)`

ARolek: path.Join() is variadic and does not like that syntax. I tried it before I ended up with this code. Check it out: https://play.golang.org/p/fSUEykW6Je3

ARolek:
@ear7h great code review. I have incorporated all your suggestions except for the last one.

server/server.go
ear7h: My bad I thought it was `func (string, string...) string`

ear7h:
LGTM

PR #642, opened 2019-09-03
Fixed URIPrefix breaking viewer and tile cache

### LINKED ISSUE #361: Use relative URLs in Tegola page
The `/` page served by Tegola uses absolute URLs, e.g. `http://localhost:8080/capabilities/bolder.json`

This is a problem because in production it's very unlikely that it will match the URL. Instead, relative URLs should be used.

ARolek:
* Fixed custom URIPrefix breaking internal viewer
* Implemented URIPrefix in autogenerated styles, which are used
by the internal viewer.
* Fixed custom URIPrefix breaking tile cache.
* Updated associated tests
* Updated viewer URLs to be relative. closes #361


ear7h:
LGTM

PR #648, opened 2019-09-30
simplify: use geom package simplify function

ear7h:
* remove usage tegola's simplify function from map.Encode
* fix bug which caused simplification to be always off
* fix typo in defaults.go
* remove maths/simplify

__Note__: ~This pr depends on https://github.com/go-spatial/geom/pull/77, or map.Encode panics.~

I left some debugging code in case changes are requested, but I'll take it out before the merge.


atlas/map.go
```diff
@@ -187,21 +191,40 @@ func (m Map) Encode(ctx context.Context, tile *slippy.Tile) ([]byte, error) {
 					}
 				}
 
+				defer func() {
```
ARolek: Can you add some code comments around this block? We migth also want to put move this to a separate function since this is debugging code and the Encode() func is already quite long.

atlas/map.go
```diff
@@ -187,21 +191,40 @@ func (m Map) Encode(ctx context.Context, tile *slippy.Tile) ([]byte, error) {
...
+
+						file, err = os.OpenFile(fname+".wkb", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+						if err == nil {
+							err = wkb.Encode(hex.NewEncoder(file), geo)
+							if err != nil {
+								log.Println("ERROR WRITING panic_dump", err)
+							}
+							file.Close()
+						}
+						// panic(r)
```
ARolek: leftover commented out code.

ARolek:
@ear7h can you rebase this to v0.11.x so the CI passes? I think this is ready for a merge.

PR #649, opened 2019-09-30
ci: install geom package in ci build

ear7h:
Since bb663be505dfab59c178595d5a5688f9650d1f7f geom is not vendored so we need to `go get` in order to get tests running.


ARolek:
@ear7h I think we need to the sqlite3 package as well

ear7h:
@ARolek the `...` at the end of `go get` installs dependencies of geom (sqlite included)

Issue #602, opened 2019-04-30
Offline tiles generation without postgis step


ingenieroariel:
A friend recently stopped using tegola and switched to Tilemaker: https://github.com/systemed/tilemaker his use case was generating offline tiles to put on a mobile app (react + mapbox).

I had initially set up Tegola for him and he developed the prototype, but when it was time to put in production he could not figure out how to get the cache from tegola to his app and tried Tilemaker, turns out it produced exactly the create and forget tiles he was expecting, starting from an osm.pbf and outputting mbtiles or a folder with tiles. 

To satisfy that use case we would need something like imposm + tegola directly to a folder we can serve with nginx (skipping postgis if possible).

pnorman:
A vector tile set normally includes data from sources that imposm can't process (e.g. natural earth). Queries are also how the layers are defined, and those queries are evaluated by postgres.

ingenieroariel:
Could naturalearth be processed with just the gdal/ogr driver? I believe
imposm bundles ogr.

-a

ARolek:
@ingenieroariel tegola can write to S3 or the file system so you could pre-generate the tileset and then serve it directly. The tiles are gzip encoded so I'm wondering if that's potentially where the issue lies. If nginx was not setting the `Content-Encoded` header correctly then the client might not know to decompress the tile prior to rendering it.

Issue #583, opened 2019-03-08
Add mbtiles as a Providers


genofire:
Use mbtiles-file from filesystem.

Do not mix it with caching to mbtiles like in #520 (where you save cache in this format)

ARolek:
I think using mbtiles as a cache provider makes sense but not so much as a data provider. Tegola data providers expect geometries to work with, not tiles. We could decode the tiles and return geometries to match the data provider interface but it sounds like your request is more around using tegola as a tile server for a pre seeded cache. 

If we supported tegola being configured with only a cache provider and no data provider then this could potentially work. We could also add a `readonly` config param for cache providers to avoid mutating the cache in this situation.

pnorman:
Readonly is what #579 is about, and I think it makes sense.

Issue #520, opened 2018-09-27
Create a cache provider for mbtile


gdey:
https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md

The MBTile spec is perfect for our cache system. We should implement this a deployment target for our cache system.

A few things to note.

The system will have to o the translation from the standard "XYZ" format to the one required by mbtiles using the [formula](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content-1) describe at: https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content-1

When setting up the cache the use will need to specify the the bounds of the map (Need to see if the cache has the ability to get this from the mapname?) and the center point. 

The Min and Max Zoom values should be those of the cache.

For this implementation we will not support [`grid` and `grid_data` tables](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#grids). 

We will assume that the system will provide us with gzipped mvt data. So, all tile data should be stored as `pbf`.

The biggest diffculty may be the `metadata` table and the [JSON field](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#vector-tileset-metadata) in it. We may need a way for the cache to get access to Atlas to get a maps metadata. We should be able to create a new mbtile file for each map.

erictheise:
Thanks for creating this issue, @gdey, it's encouraging to see your expertise shine through as you scope out what needs to be done.

pnorman:
Couple of things to keep in mind

- the mbtiles spec sets out what it needs to look like for the consumer of it, but allows different storage with views. A common way of writing mbtiles is to write x, y, z, md5 into one table, md5, tiledata into another table, and have a view that joins them
- Most mbtiles caches I've seen fail with concurrency errors when you put heavy concurrent reads and writes on them, so this is something that needs testing.

PR #654, opened 2019-11-13
ci: upgrade to go1.13

sapk:
Simple upgrade to latest go version.

I plan to work on #520 (I hope to be able to use the mbtile "offline" later) and maybe others issues somewhat related #602 #583. But before, I would suggest this unrelated change. 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,15 +37,15 @@ env:
 
 matrix:
   include:
+    - go: "1.13.x"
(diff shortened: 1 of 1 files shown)


ARolek:
@sapk LGTM. The CI is breaking due to a known issue we're working on in the geom package so disregard that issue. I'm looking forward to the mbtiles PR! Thanks for taking that one on.

PR #655, opened 2019-11-16
MBTiles cache provider

sapk:
~This is maybe early to open the PR (since the cache does nothing yet) but you can maybe already review the options and docs or make any suggestion.~

The goals is to implement #520.

This seems to works. The files generated would need to fill the json metadata to be fully compatible with specs. This could be also improved by using a separate table for the tiles data (from what other implementation seems to do) and use a view that join index table and data table to make compatible with specs.

~I still need to validate against mbtiles reader that the format is fully ok.~ seems to work with https://github.com/consbio/mbtileserver 
~And add more test for the bounds limit (that may would be better to remove at first since not mandatory for specs compatibility).~ I removed it

~I will probably create separate PR for fix found during the dev.~ done

~This would need a proper generation of the metadata json (currently string operation and a little hacky around atlas).~ done


ARolek:
@sapk is this ready for review?

sapk:
@ARolek Yes it is ready. This could definitely be improved (specially the json metadata). But it is working on an instance where I seeded the tiles. I could provide the mbtiles to validate against other client if needed.

sapk:
I have found that https://github.com/maptiler/tileserver-gl wasn't liking the json metadata and also that there is already code for vector_layers json generation so now the json generation is better (and for simplicity I store the full TileJSON not only the vector_layer fields in metadata).

sapk:
If someone want to validate the .mbtiles files: https://drive.google.com/file/d/1LmjL_TBLj5NQ6hKlY3VYR-3UNanPLvZv/view?usp=sharing
The file contains tiles for 0-6 zoom globally and under only around Paris, France.

ARolek:
@sapk I just did a quick review and I see the issue you're encountering form the metadata standpoint. Needing to fetch the metadata from `atlas` is not elegant. Maybe instead of passing through `mapName` we pass in `*atlas.Map` with the `cache.Set()` call. This will require making some minor updates to the other cache providers but it's a small change. 

I will do a deeper review with comments soon.

sapk:
Yes, I try to do something similar that on the tilejson web request which seems good enough with some shortcut but this can maybe be improved by refactoring of interface. I try first to keep the change on other part at minimum (layer interface, refactor common code under function, ...)

sapk:
I fixed the tests that I have break during rewrite of some parts and forgot to re-check. I also split some parts of this PR in #662 and #663 to ease the process and break down some part. I will update this PR if they are merged before this would reduce the change introduce here and help review.

ARolek:
@sapk can you re-open this PR and set the base to v0.12.x? The PR was closed accidentally but I still want to get this one into an upcoming release.

PR #657, opened 2019-11-17
fix: wrong cli check

sapk:
During #655, this command line check seems to fail to check lat lng validity by validating every float.


ARolek:
Good catch.

PR #661, opened 2019-11-22
Use proj library

meilinger:
Conversion to use proj library for WGS84->WebMercator and WebMercator->WGS84


proj/proj.go
```diff
@@ -215,30 +213,42 @@ func CloneGeometry(geometry geom.Geometry) (geom.Geometry, error) {
...
 	default:
-		return nil, fmt.Errorf("don't know how to convert from %v to %v.", WebMercator, SRID)
-	case WebMercator:
+		return nil, fmt.Errorf("don't know how to convert from %v to %v.", proj.WebMercator, SRID)
+	case proj.WebMercator:
 		// Instead of just returning the geometry, we are cloning it so that the user of the API can rely
 		// on the result to alway be a copy. Instead of being a reference in the on instance that it's already
 		// in the same SRID.
 
 		return CloneGeometry(geometry)
```
ARolek: @ear7h you have an [open issue](https://github.com/go-spatial/geom/issues/42) around adding `Clone()` to the geom.Geometry interface. If you implement that and port over the tegola `proj.ApplyToPoints` func in I think we can make this PR even cleaner.

atlas/map.go
```diff
@@ -23,9 +23,10 @@ import (
 	"github.com/go-spatial/geom/planar/makevalid/hitmap"
 	"github.com/go-spatial/geom/planar/simplify"
 	"github.com/go-spatial/geom/slippy"
+	"github.com/go-spatial/proj"
 	"github.com/go-spatial/tegola"
 	"github.com/go-spatial/tegola/dict"
-	"github.com/go-spatial/tegola/proj"
+	tegolaProj "github.com/go-spatial/tegola/proj"
```
ARolek: This alias is necessary because you're referring to the `proj` package constant `proj.WebMercator`. Rather than needing the alias, I propose you create package level constants in the `tegola/proj` package called `WebMercatorSRID` and `WGS84SRID` which point to the `go-spatial/proj` package. This would allow us to keep the interactions with the `go-spatial/proj` package in a single place in the codebase and drop this alias. This suggestion applies to the entire PR.

meilinger: Awesome, thanks @ARolek !  Would this suggestion also apply to the rest of the project (eg. consolidate WGS84 / WebMercator references to be internal to `github.com/go-spatial/tegola/proj` throughout) or is the main intent to drop the alias?

provider/postgis/util_internal_test.go
```diff
@@ -7,8 +7,8 @@ import (
 
 	"github.com/jackc/pgx"
 
+	"github.com/go-spatial/proj"
```
ARolek: I know it's functionally not different, but please revert this import to the `tegola/proj` package so we're consistent.

provider/test/emptycollection/provider.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"github.com/go-spatial/geom"
-	"github.com/go-spatial/tegola/proj"
+	"github.com/go-spatial/proj"
```
ARolek: I know it's functionally not different, but please revert this import to the `tegola/proj` package so we're consistent.

provider/test/provider.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	"github.com/go-spatial/geom"
-	"github.com/go-spatial/tegola/proj"
+	"github.com/go-spatial/proj"
```
ARolek: I know it's functionally not different, but please revert this import to the `tegola/proj` package so we're consistent.

PR #663, opened 2019-11-27
refactor: extract tileJSON.SetVectorLayers(m.Layers)

sapk:
Extract the refactor of tilejson generation to have a common code for #655 (which seems to fail tests).

This PR shouldn't change the logic. 
It just to break down some part of #655 to ease review process and also find what is failing on tests.


mapbox/tilejson/tilejson.go
```diff
@@ -136,3 +141,74 @@ type VectorLayer struct {
 	// Tegola supports individual layer tiles.
 	Tiles []string `json:"tiles"`
 }
+
+//SetVectorLayers fill VectorLayers from map layers
+func (tileJSON *TileJSON) SetVectorLayers(layers []atlas.Layer) {
```
ARolek: I think we should move this to atlas/map.go. The intent of the `mapbox` packages is to keep them independent of the implementation so they could potentially be moved out of the tegola codebase and stand alone. By adding references to the `atlas` package the `tilejson` package is now dependent on the `tegola` codebase.

sapk: I move the common code to the atlas package. I will also update the PR #655

sapk:
I just found a typo in the PR desc and edited it.
~This PR should change the logic.~ -> This PR shouldn't change the logic.

server/handle_map_capabilities.go
```diff
@@ -75,74 +60,13 @@ func (req HandleMapCapabilities) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
-			Extent:  4096,
-			ID:      m.Layers[i].MVTName(),
-			Name:    m.Layers[i].MVTName(),
-			MinZoom: m.Layers[i].MinZoom,
-			MaxZoom: m.Layers[i].MaxZoom,
-			Tiles: []string{
-				buildCapabilitiesURL(r, []string{"maps", req.mapName, m.Layers[i].MVTName(), "{z}/{x}/{y}.pbf"}, debugQuery),
-			},
+	//Build tiles urls
+	for i, layer := range tileJSON.VectorLayers {
```
ARolek: I see what you're doing here and why you're doing it, but it has a code smell to me. Part of what I'm hoping to achieve with this PR is positioning us to pass an atlas.Map to the Cache layer so the TileJSON can be easily generated for the MBtiles cache you built. The key issue is the dynamic generation of the HTTP host and URI path prefix. The MBTiles cache provider will need this information but I'm not sure how to best pass it down to that layer without making the interface overly complex. 

I need to think for a bit on how to adjust the Cache interface so we can achieve this while also allowing for dynamic URLS to be generated. If you're open to chatting on this, please join us on Slack: https://slack.go-spatial.org

sapk: No problem, that why I split the PR to make it more easy to review since some changes are not obvious.
I also think that it was a little "hacky" to do and try to limit it. 
I will try to pass by on slack when I have free time.

ARolek:
@sapk can you re-open this PR and set the base to v0.12.x? The PR was closed accidentally but I still want to get this one into an upcoming release.

sapk:
@ARolek I will try tomorrow.

PR #665, opened 2020-01-08
Refactored DataProvider Interface

gdey:
Cleanup and refactored DataProvider interface to support
adding differet types of data providers. This is to keep this
branch alined with tegola-postgis

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -28,8 +28,8 @@ type Config struct {
 	Webserver    Webserver `toml:"webserver"`
 	Cache        env.Dict  `toml:"cache"`
 	// Map of providers.
-	Providers []env.Dict
-	Maps      []Map
+	Providers []env.Dict `toml:"providers"`
+	Maps      []Map      `toml:"maps"`
 }
 
 type Webserver struct {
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -67,18 +67,24 @@ type MapLayer struct {
...
 	if len(plParts) != 2 {
-		return "", ErrInvalidProviderLayerName{ProviderLayerName: string(ml.ProviderLayer)}
+		return "", "", ErrInvalidProviderLayerName{ProviderLayerName: string(ml.ProviderLayer)}
 	}
 
-	return plParts[1], nil
+	return plParts[0], plParts[1], nil
+}
+
+// GetName helper to get the name we care about.
```
ARolek: Let's adjust this comment to:

```go
// GetName will return the user-defined Layer name from the config, 
// or if not defined will return the name of the layer associated with the provider
```

config/config.go
```diff
@@ -67,18 +67,24 @@ type MapLayer struct {
 	DontClip env.Bool `toml:"dont_clip"`
 }
 
-// GetName helper to get the name we care about.
-func (ml MapLayer) GetName() (string, error) {
-	if ml.Name != "" {
-		return string(ml.Name), nil
-	}
+// GetProviderLayerName returns the provider and layer names
+func (ml MapLayer) GetProviderLayerName() (string, string, error) {
```
ARolek: Drop `Get` from method prefix.

PR #687, opened 2020-05-21
Migrate mvt providers

gdey:
This brings in the mvt_providers features.

### DIFF
--- a/cmd/internal/register/errors.go
+++ b/cmd/internal/register/errors.go
@@ -1,0 +1,55 @@
+package register
+
+import "fmt"
+
+// ErrProviderLayerInvalid should be returned when an invalid Provider layer for a map is given
+type ErrProviderLayerInvalid struct {
+	ProviderLayer string
+	Map           string
+}
+
+func (e ErrProviderLayerInvalid) Error() string {
+	return fmt.Sprintf("invalid provider layer (%v) for map (%v)", e.ProviderLayer, e.Map)
+}
+
+// ErrProviderNotFound when the requested provider is not a known provider
+type ErrProviderNotFound struct {
+	Provider string
+}
+
+func (e ErrProviderNotFound) Error() string {
+	return fmt.Sprintf("provider (%v) not defined", e.Provider)
+}
+
+// ErrProviderLayerNotRegistered should be returned when the requested provider was not registered into the system.
+type ErrProviderLayerNotRegistered struct {
+	MapName       string
+	ProviderLayer string
+	Provider      string
+}
+
+func (e ErrProviderLayerNotRegistered) Error() string {
+	return fmt.Sprintf("map (%v) 'provider_layer' (%v) is not registered with provider (%v)", e.MapName, e.ProviderLayer, e.Provider)
+}
+
+// ErrFetchingLayerInfo wraps an error when attempting to obtain layer information for a provider
+type ErrFetchingLayerInfo struct {
+	Provider string
+	Err      error
+}
+
+func (e ErrFetchingLayerInfo) Unwrap() error { return e.Err }
+func (e ErrFetchingLayerInfo) Error() string {
+	return fmt.Sprintf("error fetching layer info from provider (%v): %v", e.Provider, e.Err)
+}
+
+// ErrDefaultTagsInvalid should be returned when the type of defaultTags is not an acceptable type.
+type ErrDefaultTagsInvalid struct {
+	ProviderLayer string
+	Err           error
+}
+
+func (e ErrDefaultTagsInvalid) Unwrap() error { return e.Err }
+func (e ErrDefaultTagsInvalid) Error() string {
+	return fmt.Sprintf("'default_tags' for 'provider_layer' (%v) should be a TOML table", e.ProviderLayer)
+}
--- a/cmd/internal/register/maps.go
+++ b/cmd/internal/register/maps.go
@@ -1,173 +1,160 @@
 package register
 
 import (
-	"fmt"
 	"html"
-	"strings"
 
 	"github.com/go-spatial/geom"
-	"github.com/go-spatial/tegola"
 	"github.com/go-spatial/tegola/atlas"
 	"github.com/go-spatial/tegola/config"
+	"github.com/go-spatial/tegola/mvtprovider"
 	"github.com/go-spatial/tegola/provider"
 )
 
-type ErrProviderLayerInvalid struct {
-	ProviderLayer string
-	Map           string
+func webMercatorMapFromConfigMap(cfg config.Map) (newMap atlas.Map) {
+	newMap = atlas.NewWebMercatorMap(string(cfg.Name))
+	newMap.Attribution = html.EscapeString(string(cfg.Attribution))
+
+	// convert from env package
+	for i, v := range cfg.Center {
+		newMap.Center[i] = float64(v)
+	}
+
+	if len(cfg.Bounds) == 4 {
+		newMap.Bounds = geom.NewExtent(
+			[2]float64{float64(cfg.Bounds[0]), float64(cfg.Bounds[1])},
+			[2]float64{float64(cfg.Bounds[2]), float64(cfg.Bounds[3])},
+		)
+	}
+
+	if cfg.TileBuffer != nil {
+		newMap.TileBuffer = uint64(*cfg.TileBuffer)
+	}
+	return newMap
+
 }
 
-func (e ErrProviderLayerInvalid) Error() string {
-	return fmt.Sprintf("invalid provider layer (%v) for map (%v)", e.ProviderLayer, e.Map)
+func layerInfosFindByName(infos []provider.LayerInfo, name string) provider.LayerInfo {
+	if len(infos) == 0 {
+		return nil
+	}
+	for i := range infos {
+		if infos[i].Name() == name {
+			return infos[i]
+		}
+	}
+	return nil
 }
 
-type ErrProviderNotFound struct {
-	Provider string
+func atlasLayerFromConfigLayer(cfg *config.MapLayer, mapName string, layerProvider provider.Layerer) (layer atlas.Layer, err error) {
+	var (
+		// providerLayer is primary used for error reporting.
+		providerLayer = string(cfg.ProviderLayer)
+		ok            bool
+	)
+	// read the provider's layer names
+	// don't care about the error.
+	providerName, layerName, _ := cfg.ProviderLayerName()
+	layerInfos, err := layerProvider.Layers()
+	if err != nil {
+		return layer, ErrFetchingLayerInfo{
+			Provider: providerName,
+			Err:      err,
+		}
+	}
+	layerInfo := layerInfosFindByName(layerInfos, layerName)
+	if layerInfo == nil {
+		return layer, ErrProviderLayerNotRegistered{
+			MapName:       mapName,
+			ProviderLayer: providerLayer,
+			Provider:      providerName,
+		}
+	}
+	layer.GeomType = layerInfo.GeomType()
+
+	if cfg.DefaultTags != nil {
+		if layer.DefaultTags, ok = cfg.DefaultTags.(map[string]interface{}); !ok {
+			return layer, ErrDefaultTagsInvalid{
+				ProviderLayer: providerLayer,
+			}
+		}
+	}
+
+	// if layerProvider is not a provider.Tiler this will return nil, so
+	// no need to check ok, as nil is what we want here.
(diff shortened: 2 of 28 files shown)


gdey:
@ARolek I tried to test this, but something is wrong with my Postgres configuration. Everything runs, but Postgres complains when producing the SQL with an error stating ST_AsMVT does not exist.

gdey:
Should we try and implement this issue, for this pull request?

https://github.com/go-spatial/tegola-postgis/issues/4

mvtprovider/layer.go
```diff
@@ -0,0 +1,11 @@
+package mvtprovider
+
+// layer holds information about a query.
```
ARolek, line 3: Capitalize "layer"

mvtprovider/postgis/README.md
```diff
@@ -0,0 +1,127 @@
...
+### Connection Properties
+
+- `name` (string): [Required] provider name is referenced from map layers. please note that when referencing an mvt_provider form a map layer the provider name must be prexied with `mvt_`. See example config below.
+- `type` (string): [Required] the type of data provider. must be "postgis" to use this data provider
+- `host` (string): [Required] PostGIS database host
+- `port` (int): [Required] PostGIS database port (required)
+- `database` (string): [Required] PostGIS database name
+- `user` (string): [Required] PostGIS database user
+- `password` (string): [Required] PostGIS database password
+- `srid` (int): [Optional] The default SRID for the provider. Defaults to WebMercator (3857) but also supports WGS84 (4326)
```
ARolek, line 25: Does this apply? For MVTProviders we have no control over reprojecting the data (for now)

mvtprovider/postgis/README.md
```diff
@@ -0,0 +1,127 @@
...
+# this table uses "geom" for the geometry_fieldname and "gid" for the id_fieldname so they don't need to be configured
+sql = "SELECT ST_AsMVTGeom(geom,!BBOX!) AS geom, gid FROM gis.landuse WHERE geom && !BBOX!"
+```
+
+### Provider Layers Properties
+
+- `name` (string): [Required] the name of the layer. This is used to reference this layer from map layers.
+- `geometry_fieldname` (string): [Optional] the name of the filed which contains the geometry for the feature. defaults to `geom`.
+- `id_fieldname` (string): [Optional] the name of the feature id field. defaults to `gid`.
+- `srid` (int): [Optional] the SRID of the layer. Supports `3857` (WebMercator) or `4326` (WGS84).
```
ARolek, line 43: Does this apply? For MVTProviders we have no control over reprojecting the data (for now)

mvtprovider/postgis/README.md
```diff
@@ -0,0 +1,127 @@
...
+  - `LAYER_SQL:EXECUTE_SQL`: print `LAYER_SQL` and `EXECUTE_SQL`.
+
+Example:
+
+```
+$ TEGOLA_SQL_DEBUG=LAYER_SQL tegola serve --config=/path/to/conf.toml
+```
+
+## Testing
+Testing is designed to work against a live PostGIS database. To see how to set up a database check this [ci script](https://github.com/go-spatial/tegola/blob/master/ci/config_postgis.sh). To run the PostGIS tests, the following environment variables need to be set:
```
ARolek, line 107: The referenced CI script is no longer in the repo.

mvtprovider/postgis/README.md
```diff
@@ -0,0 +1,127 @@
...
+  sql = "SELECT ST_AsMVTGeom(geom,!BBOX!) AS geom, gid FROM gis.landuse WHERE geom && !BBOX!"
+
+[[maps]]
+name = "cities"
+center = [-90.2,38.6,3.0]  # where to center of the map (lon, lat, zoom)
+
+    [[maps.layers]]
+    name = "landuse"
+    # note the mvt_ prefix on the name of the provider.
+    provider_layer = "mvt_test_postgis.landuse"
```
ARolek, line 85: I still think this is a confusing gotcha for the end-user. Ideally we could have the `type` use the `mvt_` prefix, but that does complicate the checking in another way. Just something we can think about from a UX perspective.

provider/errors.go
```diff
@@ -17,3 +18,34 @@ type ErrUnableToConvertFeatureID struct {
...
+
+func (err ErrProviderAlreadyExists) Error() string {
+	return fmt.Sprintf("provider %s already exists", err.Name)
+}
+
+// ErrUnknownProvider is retured when no providers are registered or requested
+// provider is not registered
+type ErrUnknownProvider struct {
+	Name               string
+	KnownProvidersFunc func() []string
```
ARolek, line 36: Seems strange to me to embed a function like this into an error. Maybe consider changing this to `KnownProviders []string` and setting the value at the time of the error.

provider/postgis/postgis.go
```diff
@@ -606,6 +603,64 @@ func (p Provider) TileFeatures(ctx context.Context, layer string, tile provider.
...
+	if debugExecuteSQL {
+		log.Printf("%s:%s: %v", EnvSQLDebugName, EnvSQLDebugExecute, fsql)
+	}
+	err = p.pool.QueryRow(fsql).Scan(&data)
+	if debugExecuteSQL {
+		log.Printf("%s:%s: %v", EnvSQLDebugName, EnvSQLDebugExecute, fsql)
+		if err != nil {
+			log.Printf("%s:%s: returned %v bytes", EnvSQLDebugName, EnvSQLDebugExecute, len(data.Bytes))
+		} else {
+			log.Printf("%s:%s: returned error %v", EnvSQLDebugName, EnvSQLDebugExecute, err)
```
ARolek, line 653: I think these logs are inversed.

provider/errors.go
gdey: The reason for this, is to lazy load the known providers, instead of have a list passed around the whole time.

ARolek: But why not just call Drivers() right when the error is created and put the list of known drivers at that time in the error? I know it's basically an impossible case right now, but technically the list of drivers could change between the time the error is generated and the Error() method is called.

gdey: That's a fair point

cmd/internal/register/maps.go
```diff
@@ -41,133 +40,171 @@ func (e ErrProviderLayerNotRegistered) Error() string {
...
-			if !ok {
-				return ErrProviderNotFound{providerLayer[0]}
+	if cfg.DefaultTags != nil {
+		if layer.DefaultTags, ok = cfg.DefaultTags.(map[string]interface{}); !ok {
+			return layer, ErrDefaultTagsInvalid{
+				ProviderLayer: providerLayer,
 			}
+		}
+	}
+	layer.Provider, _ = layerProvider.(provider.Tiler)
```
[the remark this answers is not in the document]
gdey: yes; I'll put a comment.

mvtprovider/postgis/README.md
gdey: #688 covers this.

gdey:
@ARolek could use your help on the MVT_Provider test. For some reason the database keeps returning 0 bytes to me. Not sure what's going on. I check it in PGAdmin 4 and it also returns me zero bytes; when using ST_AsMVT, but ST_AsGeom sub-select returns the right number of geoms.

mvtprovider/postgis/README.md
```diff
@@ -0,0 +1,131 @@
...
+  - `!SCALE_DENOMINATOR!` - [Optional] scale denominator, assuming 90.7 DPI (i.e. 0.28mm pixel size)
+  - `!PIXEL_WIDTH!` - [Optional] the pixel width in meters, assuming 256x256 tiles
+  - `!PIXEL_HEIGHT!` - [Optional] the pixel height in meters, assuming 256x256 tiles
+  - `!ID_FIELD!` - [Optional] the id field name
+  - `!GEOM_FIELD!` - [Optional] the geom field name
+  - `!GEOM_TYPE!` - [Optional] the geom type if defined otherwise ""
+
+## Example mvt_provider and map config
+
+**Important**: When referencing the `provider` in the `map` section of the config, you MUST prepend `mvt_` to the `provider_layer` value. This indicates to tegola that the provider is an MVT provider so tegola knows which provider section to perform the lookup.
```
ear7h, line 64: Sorry if this has been talked about in person, or if I'm missing something, but it doesn't seem like the prefix is needed. When initializing `Map` both the `provider` and `mvtprovider` maps are checked:

https://github.com/go-spatial/tegola/blob/303488246a833a65a77e5fa0d0ee8029a187f0b3/cmd/internal/register/maps.go#L115-L118

Though, there needs to be an extra loop to check naming collisions like:
```go
// check name collisions between regular providers and mvtproviders
for k := range providers {
    if _, ok := mvtProviders[k]; ok {
        return provider.ErrProviderAlreadyExists { 
            Name: k,
        }
    }
}
```
before the following lines
https://github.com/go-spatial/tegola/blob/303488246a833a65a77e5fa0d0ee8029a187f0b3/cmd/tegola/cmd/root.go#L94
https://github.com/go-spatial/tegola/blob/303488246a833a65a77e5fa0d0ee8029a187f0b3/cmd/tegola_lambda/main.go#L86

cmd/internal/register/maps.go
```diff
@@ -1,173 +1,165 @@
...
+	return layer, nil
+}
 
-		// convert from env package
-		centerArr := [3]float64{}
-		for i, v := range m.Center {
-			centerArr[i] = float64(v)
+func selectProvider(name string, mapName string, newMap *atlas.Map, providers map[string]provider.Tiler, mvtProviders map[string]mvtprovider.Tiler) (provider.Layerer, error) {
+	if newMap.HasMVTProvider() {
+		// okay we need to make sure the provider name matchs up.
```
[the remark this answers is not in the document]
gdey: hehe, Yeah this was a more of comment to me, when I was originally building up this code.

cmd/internal/register/errors.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func (e ErrProviderLayerNotRegistered) Error() string {
+	return fmt.Sprintf("map (%v) 'provider_layer' (%v) is not registered with provider (%v)", e.MapName, e.ProviderLayer, e.Provider)
+}
+
+// ErrFetchingLayerInfo wraps an error when attempting to obtain layer information for a provider
+type ErrFetchingLayerInfo struct {
+	Provider string
+	Err      error
```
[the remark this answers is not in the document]
gdey: Error here is export to allow Creating the ErrFetchingLayerInfo.  You have a point. I need to think about if it makes sense to make it the only private variable in this struct. I like keeping the ErrStructs pretty dead simple.

cmd/internal/register/errors.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ErrDefaultTagsInvalid should be returned when the type of defaultTags is not an acceptable type.
+type ErrDefaultTagsInvalid struct {
+	ProviderLayer string
+	Err           error
+}
+
+func (e ErrDefaultTagsInvalid) Unwrap() error { return e.Err }
+func (e ErrDefaultTagsInvalid) Error() string {
+	return fmt.Sprintf("'default_tags' for 'provider_layer' (%v) should be a TOML table", e.ProviderLayer)
```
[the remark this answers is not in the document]
gdey: I had it initially with the provider_layer name, and found that it was hard to debug. Putting in the context help debug the config quite a bit more.

cmd/internal/register/maps_test.go
```diff
@@ -32,17 +34,24 @@ func TestMaps(t *testing.T) {
...
 		}
+
+		err = register.Maps(&tc.atlas, tc.maps, providers, mvtProviders)
+		if !errors.Is(err, tc.expectedErr) {
+			t.Errorf("invalid error, expected %v got %v", tc.expectedErr, err)
+		}
+		return
 	}
 
 	tests := map[string]tcase{
```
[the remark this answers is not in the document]
gdey: yes, I just ran out of time for those. Just create #695 to track this. Thanks.

mvtprovider/postgis/README.md
gdey: It's because of the config; if I remember correctly. And registering of the provider. We will eventually move to use the `type` field instead of having different sections. #688 tracks this.

cmd/internal/register/errors.go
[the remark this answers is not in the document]
gdey: I add this after I had a hard time debugging config files; because got confusing which layer/provider in the config was causing the issue. This error ends up getting thrown pretty far in the initialization process.

Issue #610, opened 2019-07-02
[Feature Request] Add support for filtering layer dynamically


aaj013:
I haven't seen any options right now within Tegola to pass on parameters from frontend to DB for filtering the input dataset. Is there any updates on that?

ARolek:
@aaj013 it's not currently on our roadmap. Could you give a more concrete example on how you would want this to work from the frontend? It's an interesting idea though I see it conflicting with concepts like tile caching, but I'm open to exploring the possibility.

aaj013:
I see the same functionality in another vector tile server [Martin](https://github.com/urbica/martin) passing on parameters from client side to plpgsql (https://github.com/urbica/martin#function-sources). Wouldn't it be nice to integrate that feature?

tomass:
One possible use case:

There is a tourism map/website allowing you to choose a number of POI types to display from a list of ~50+. It then uses a bbox strategy to request a list of poi's in a viewport (bbox), something like _poi.php?bbox=12,23,45,56&type=ABC_ (where A, B and C are types of POI to be displayed). This works with OpenLayers.

While trying to migrate to mapbox-gl-js, there is a problem: bbox strategy is not available in mapbox-gl-js.
Putting all 50+ categories of POI's into one source of vector tiles is not nice, as there is a requirement to display POI's even on small scales (putting all POI's on small scales would make vtiles too large).

Having as many vector tile sources/layers as POI types (50+) would be too cumbersome/slow when a lot of POI types is selected. And map style file would have to include all 50+ types separately.

Solution could be to use similar request url but with parameter, say: _15/78/456.pbf?type=ABC_
Argument "type" with value "ABC" could be passed to Tegola so that it does a one-time query, creates a pbf, returns it and does not save in a cache. Query could be written like this:

```
select ...
from ...
where instr(:type, field_type) > 0
```

P.S. This is just a theoretical use case. I have not tested if this would be usable in mapbox-gl-js.

ARolek:
@aaj013 one thing about martin is that it only supports PostGIS. Tegola has more than one provider type so I'm always trying to think of how we can make the implementation provider agnostic. I have not dug into Martin too much, but do you know if it supports a tile cache along with the query feature? I think that would be an interesting challenge to tackle.

@tomass this is an interesting use case and seems to be very specific. I still think we're going to run into issues with a tile cache, but I guess in this use case there would be no tile cache. I'm not opposed to this feature, just trying to understand the use case more.

[...]

PR #700, opened 2020-07-09
Merge mvt_providers and providers into one section.

### LINKED ISSUE #688: Merge mvt_provider and providers into single providers section in config.
This is a copy of https://github.com/go-spatial/tegola-postgis/issues/4

with #687 we will have the mvt_provider experiment in the mainline, and so should track currently known issues with it.

Should we remove mvt_provider, and have the provider register as an MVT type provider. 

i.e.
```toml
[[providers]]
name = "test_postgis"       # provider name is referenced from map layers (required)
type = "mvt_postgis"            # the type of data provider must be "postgis" for this data provider (required)
host = "localhost"          # PostGIS database host (required)
port = 5432                 # PostGIS database port (required)
database = "tegola"         # PostGIS database name (required)
user = "tegola"             # PostGIS database user (required)
password = ""               # PostGIS database password (required)
```

There are some things to note about this approach.

1. The provider needs to state that it's an MVT provider (this could be done by look to see if it supports the mvt_provider interface)

2. Documentation can be confusing as mvt providers can not conflate with other providers.

# Errors

If a map has an mvt provider, it can be the only one and the system will error during configuration stating that the mvt_provider can be the only provider.

gdey:
Merge mvt_providers and providers sections of the config.

We will now use the type param to figure out the mvt_providers and providers.

fixes #688


provider/provider.go
```diff
@@ -10,6 +10,37 @@ import (
 	"github.com/go-spatial/tegola/internal/log"
 )
 
+type providerType uint8
+
+const (
```
ear7h, line 15: I think only the `providerType` constants should be defined, and be bit flags. And the filters are implemented with a function:
```go
func ProviderFilterInclude(filters ...providerType) providerFilter {
    ret := uint8(0)
    for _, v := range filters {
        ret |= uint8(v)
    }
    return providerFilter(ret)
}
```

Having two sets of constants that essentially refer to the same information could be confusing

provider/provider.go
```diff
@@ -59,33 +90,46 @@ type Tiler interface {
...
-	GeomType() geom.Geometry
-	SRID() uint64
+func (tu TilerUnion) Layers() ([]LayerInfo, error) {
+	if tu.Std != nil {
+		return tu.Std.Layers()
+	}
+	if tu.Mvt != nil {
+		return tu.Mvt.Layers()
+	}
+	return nil, ErrNilInitFunc
```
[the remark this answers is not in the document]
gdey: It could be. It could, also, be a panic. Was not sure which of the three options made the most sense. When with NilInitFunc because the only way one should get here is if someone added a new provider type and forgot to register the provider type in this area of the code, but maybe panic is better in that sense? I went with ErrNilFunc because that's the error you get from the init functions, and it's not panic. And having another error case got a case that should never occur did not seem right. (But not I'm leaning more towards the panic)

cmd/tegola/cmd/root.go
```diff
@@ -78,20 +78,22 @@ func initConfig(configFile string, cacheRequired bool) (err error) {
...
 
-	mvtProviders, err := register.MVTProviders(mvtProvArr)
-	if err != nil {
-		return fmt.Errorf("could not register mvt_providers: %v", err)
-	}
+		mvtProviders, err := register.MVTProviders(mvtProvArr)
+		if err != nil {
+			return fmt.Errorf("could not register mvt_providers: %v", err)
+		}
+	*/
```
ARolek, line 93: Can this be removed?

config/config.go
```diff
@@ -92,17 +98,42 @@ func (ml MapLayer) GetName() (string, error) {
...
+	drivers := make(map[string]int)
+	for _, name := range provider.Drivers(provider.ProviderFilterStd) {
+		drivers[name] = provider.ProviderFilterStd
+		knownTypes = append(knownTypes, name)
+	}
+	for _, name := range provider.Drivers(provider.ProviderFilterMVT) {
+		drivers[name] = provider.ProviderFilterMVT
+		knownTypes = append(knownTypes, name)
+	}
+	// mvtproviders maps a known provider name to weather that provider is
```
ARolek, line 111: weather -> whether

config/errors.go
```diff
@@ -91,3 +103,50 @@ type ErrInvalidURIPrefix string
 func (e ErrInvalidURIPrefix) Error() string {
 	return fmt.Sprintf("config: invalid uri_prefix (%v). uri_prefix must start with a forward slash '/' ", string(e))
 }
+
+// ErrUnknownProviderType is return when the config contains a provider type that has not been registered
+type ErrUnknownProviderType struct {
+	Name           string // name of the entry in the config
+	Type           string // name of the data provider
```
ARolek, line 110: name -> type

config/errors.go
```diff
@@ -91,3 +103,50 @@ type ErrInvalidURIPrefix string
 func (e ErrInvalidURIPrefix) Error() string {
 	return fmt.Sprintf("config: invalid uri_prefix (%v). uri_prefix must start with a forward slash '/' ", string(e))
 }
+
+// ErrUnknownProviderType is return when the config contains a provider type that has not been registered
```
ARolek, line 107: return -> returned

config/errors.go
```diff
@@ -91,3 +103,50 @@ type ErrInvalidURIPrefix string
...
+	Name           string // name of the entry in the config
+	Type           string // name of the data provider
+	KnownProviders []string
+}
+
+func (e ErrUnknownProviderType) Error() string {
+	return fmt.Sprintf("config: invalid type (%s) for provider %s; known providers are: %v", e.Name, e.Type, strings.Join(e.KnownProviders, ","))
+}
+
+// Is returns weather the error is of type ErrUnknownProviderType, only checking the Type value.
```
ARolek, line 118: weather -> whether

config/errors.go
```diff
@@ -91,3 +103,50 @@ type ErrInvalidURIPrefix string
...
+func (e ErrUnknownProviderType) Is(err error) bool {
+	err1, ok := err.(ErrUnknownProviderType)
+	if !ok {
+		return false
+	}
+	return err1.Type == e.Type
+}
```
ARolek, line 125: If the user is looking for a type comparison, shouldn't they use `As()`? It seems to me this is overriding `Is()` to make it function like `As()`

provider/errors.go
```diff
@@ -49,3 +50,32 @@ func (err ErrUnknownProvider) Error() string {
...
+	Name           string
+	Type           providerType
+	KnownProviders []string
+}
+
+func (err ErrInvalidProviderType) Error() string {
+	var errStr strings.Builder
+	fmt.Fprintf(&errStr, "provider '%v' is not of type %v", err.Name, err.Type)
+	if len(err.KnownProviders) != 0 {
+		fmt.Fprintf(&errStr, ", known providers of type (%v):", err.Type)
```
ARolek, line 66: Is this log intentionally left in?

provider/provider.go
```diff
@@ -10,6 +10,37 @@ import (
...
+
+type providerFilter uint8
+
+const (
+	ProviderFilterNone providerFilter = 0
+	ProviderFilterStd                 = 1 << iota
+	ProviderFilterMVT
+
+	ProviderFilterAll = ProviderFilterStd | ProviderFilterMVT
+)
```
ARolek, line 42: There's [another issue](https://github.com/go-spatial/tegola/issues/610) around filtering features at the provider layer. I'm thinking we might run into a naming collision with this filter concept and the data filter concept. Maybe we make these const private if possible, or be prepared to potentially change the name for data filtering.

provider/provider.go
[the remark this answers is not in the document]
ARolek: I don't know if a panic would be too user friendly. I suspect this error will most commonly be encountered when someone fat fingers a provider type, right? For example, putting in `mvt_postgres` instead of `mvt_postgis`. If that's the case, we could report an error to guide the user to resolution.

provider/provider.go
```diff
@@ -127,10 +232,40 @@ func For(name string, config dict.Dicter) (Tiler, error) {
...
+		}
+	}
 
 	return p.init(config)
 }
 
+// MVTFor function returns a configured mvt provider of the given type, provided the correct config map.
+func MVTFor(name string, config dict.Dicter) (MVTTiler, error) {
+	if providers == nil {
+		return nil, ErrUnknownProvider{}
```
ARolek, line 250: Should this error be expanded to `ErrUnknownProvider{KnownProviders: Drivers(ProviderFilterMvt)}`? If not, maybe review the `STDFor` function so these two scenarios align.

config/errors.go
gdey: It's not quite. As type as to be the same, it's not just that it's the same error type, but then the `Type` field has to be the same as well. 

`ErrUnknownProviderType{ Type: "foo" }` is different than `ErrUnknownProviderType{ Type: "bar" }`, the reason for the overload, is because the KnownProvider and Name fields are not important, and are only there to help with the error message.

ARolek: Yep, now that I re-read it I see what you're doing. e.Type is a property not the actual type of the struct.

provider/provider.go
gdey: I think I can get rid of the STDFor and MVTFor, but you are correct it should be filtered.

provider/provider.go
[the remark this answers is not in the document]
gdey: In reply to:
https://github.com/go-spatial/tegola/pull/700#discussion_r453957361


No, that error would be caught by the Config Validation long before it reaches here.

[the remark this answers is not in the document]
gdey: > I don't know if a panic would be too user friendly. I suspect this error will most commonly be encountered when someone fat fingers a provider type, right? For example, putting in `mvt_postgres` instead of `mvt_postgis`. If that's the case, we could report an error to guide the user to resolution.

No this would be caught by the config validation code earlier.

ARolek:
I finally had some time to test this out and it's working great. Please rebase onto v0.12.x prior to merging.

config/errors.go
gdey: The comment is indicating that `Type` is really the name of the data provider
And that Name here is the anme of the entry in the config. 
I'll change the comments to the following:
// Name is the name of the entry in the config
// Type is the name of the data provider

provider/errors.go
gdey: Nope

gdey: Looking at this again. That is not a log statement, but building up the error string to return.
It's looks like a log statment, I know. I made that error as well.

ARolek:
Looks good! Just clean up the commits and then let's get this into v0.12.x!

PR #704, opened 2020-07-21
provider/postgis: fix dropped errors

alrs:
This fixes two dropped `err` variables in `provider/postgis`.


ARolek:
Great catches. Thank you!

PR #706, opened 2020-07-27
maths/webmercator: fix no-op NaN check

alrs:
This fixes the check on `PLonToX()` to truly check if `lon` is `NaN`.


gdey:
Thank you!

PR #708, opened 2020-07-29
Added mapl and arolek to codeowners for ui

gdey:
Updated codeowners so that pr are tagged correctly for reviewers


ARolek:
LGTM

PR #711, opened 2020-08-12
Normalize table tests

gdey:


mvtprovider/postgis/doc.go
```diff
@@ -1,6 +1,5 @@
 // Package postgis is a placeholder for the postgis database.
 // The postgis provider is provided by the provider.Postgis provider
 // please consult that for information on how to use it.
-// please note that the provider will be called mvt_postgis in the
-// map layers
+// please note that the provider type will be mvt_postgis
 package postgis
```
ARolek, line 5: should we maybe just name this package `mvt_postgis`?

gdey: Generally you don't want "_" in the name of pacakges. Also, it's already in the mvtprovider package, and it seems redundant to have the mvt there again.

server/handle_map_style_test.go
```diff
@@ -177,6 +177,6 @@ func TestHandleMapStyleCORS(t *testing.T) {
 
 	for name, tc := range tests {
 		tc := tc
```
gdey, line 179: Need to remove this.

server/handle_capabilities_test.go
```diff
@@ -322,6 +323,6 @@ func TestHandleCapabilitiesCORS(t *testing.T) {
 
 	for name, tc := range tests {
 		tc := tc
```
gdey, line 325: remove this

PR #715, opened 2020-08-15
provider/postgis: close query after err check

alrs:
This changes a spot in `provider/postgis` where a deferred `Close` to a database query was happening before the `err` check.

### DIFF
--- a/provider/postgis/util_internal_test.go
+++ b/provider/postgis/util_internal_test.go
@@ -140,11 +140,11 @@ func TestDecipherFields(t *testing.T) {
 	fn := func(tc tcase) func(t *testing.T) {
 		return func(t *testing.T) {
 			rows, err := conn.Query(tc.sql)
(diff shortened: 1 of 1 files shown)


provider/postgis/util_internal_test.go
```diff
@@ -140,11 +140,11 @@ func TestDecipherFields(t *testing.T) {
 	fn := func(tc tcase) func(t *testing.T) {
 		return func(t *testing.T) {
 			rows, err := conn.Query(tc.sql)
-			defer rows.Close()
 			if err != nil {
 				t.Errorf("Error performing query: %v", err)
 				return
 			}
+			defer rows.Close()
```
ARolek, line 147: Do we know that if an error is returned rows.Close() does not need to be called? This is in a test case, so it's not as important, but reading this it looks like we could get into a situation where rows are not being closed when an error condition is hit.

PR #723, opened 2020-10-23
cache/s3: replace deprecated S3 function

alrs:
This replaces the deprecated `session.New()`.

`go doc github.com/aws/aws-sdk-go/aws/session.New`
>Deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.


gdey (approved):
LGTM

PR #729, opened 2020-12-11
Version bumps in Dockerfile

ARolek:
* Bump Go to 1.5.6
* Bump alpine linux to 3.12
* Bump musl-dev to 1.1.24r10
* Bump gcc to 9.3.0-r2
* Fixed Go linker change that happened in Go 1.15


gdey (approved):
LGTM

Issue #736, opened 2021-02-12
MacOS release builds don't include support for GeoPackage providers


flother:
The [GeoPackage section of the configuration docs](https://tegola.io/documentation/configuration/#geopackage) says:

> The GeoPackage provider requires that Tegola is built with CGO. Prebuilt CGO binaries can be found [here](https://github.com/go-spatial/tegola/releases) and are indicated with the suffix `_cgo`.

The `_cgo` releases haven't been published since [v0.10.4](https://github.com/go-spatial/tegola/releases/tag/v0.10.4), when releases started to be built with GitHub Actions instead of Travis CI ([commit 70d23e20](https://github.com/go-spatial/tegola/commit/70d23e20b60dbefd75cf9f845eca0912df804dab)). That commit says that "all builds now leverage CGO". My first thought was that that means all release builds now support GeoPackage providers by default, and that the sentence in the config docs should be removed.

But it doesn't seem like this is the case. If you include a GeoPackage provider in `config.toml`, the release builds will complain:

    Error: config: invalid type (foobar) for provider gpkg; known providers are: debug,postgis,mvt_postgis

To reproduce this, create a `config.toml` containing only this (the referenced file, `foobar.gpkg`, doesn't need to exist):

```toml
[[providers]]
name = "foobar"
type = "gpkg"
filepath = "foobar.gpkg"
```

Then download and run a release build of Tegola (I'm using a Mac so I used the Darwin build):

```sh
$ curl --location https://github.com/go-spatial/tegola/releases/download/v0.12.1/tegola_darwin_amd64.zip -o tegola.zip
$ unzip tegola.zip
$ ./tegola version
v0.12.1
$ ./tegola serve
2021-02-12 11:07:38 [INFO] root.go:61: Loading config file: config.toml
2021-02-12 11:07:38 [INFO] config.go:305: loading local config (config.toml)
Error: config: invalid type (foobar) for provider gpkg; known providers are: debug,postgis,mvt_postgis
Usage:
  tegola serve [flags]

Aliases:
  serve, server

Flags:
  -h, --help          help for serve
  -n, --no-cache      turn off the cache
  -p, --port string   port to bind tile server to (default ":8080")

Global Flags:
      --config string   path to config file (default "config.toml")

config: invalid type (foobar) for provider gpkg; known providers are: debug,postgis,mvt_postgis
```

[...]

PR #738, opened 2021-02-13
Separate Linux and MacOS CI release build jobs

flother:
Compiling MacOS release builds on Linux — as the GitHub Actions CI job does now — means cgo isn't enabled, which means the go-sqlite3 package isn't compiled, which means no GeoPackage support in the MacOS builds.

This pull request separates the jobs so that Linux builds are compiled on the `ubuntu-latest` runner while MacOS builds are compiled on the `macos-latest` runner. This means no cross-compilation is required and cgo can run happily.

Resolves go-spatial/tegola#736

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -27,13 +27,10 @@ jobs:
         name: version
         path: ${{ github.workspace }}/version.txt
 
-  build_nix:
-    name: Build for Linux, MacOS and AWS Lambda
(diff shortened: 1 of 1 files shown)


ARolek (approved):
LGTM! Could you please squash your commits into a single one?

Thanks for the contribution!

flother:
> Could you please squash your commits into a single one?

Done!

gdey (approved):
LGTM

PR #744, opened 2021-03-02
Fix SQL parsing for MVT provider

johngian:
* Make ID field optional
* Fix geometry sniffing for MVT queries
* Fix handling MVT provider SQL when using (SELECT ...) as alias queries

### DIFF
--- a/provider/postgis/util.go
+++ b/provider/postgis/util.go
@@ -16,8 +16,13 @@ import (
 	"github.com/jackc/pgx/pgtype"
 )
 
+// isMVT will return true if the provider is MVT based
+func isMVT(providerType string) bool {
+	return providerType == "mvt_postgis"
+}
+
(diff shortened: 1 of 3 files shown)


ARolek:
@johngian it looks like this PR is failing the test suite: https://github.com/go-spatial/tegola/pull/744/checks?check_run_id=2015097575. I can help look into this if that would be helpful.

gdey (approved):
LGTM

provider/postgis/util.go
```diff
@@ -16,8 +16,13 @@ import (
 	"github.com/jackc/pgx/pgtype"
 )
 
+// isMVT will return true if the provider is MVT based
+func isMVT(providerType string) bool {
+	return providerType == "mvt_postgis"
```
ARolek, line 21: minor nit, but since we're using `"mvt_postgis"` in more than 1 place (register.go and here), we should probably move it to a const.

Issue #719, opened 2020-09-09
Support Minio S3-Cache


richardbischof:
Dear community,

today tegola can make use of s3 as a cache. For local testing and deployment in an fully open source environments, I want to use [https://github.com/minio/minio](https://github.com/minio/minio) as my object store of choice. Minio is an open source object store, which is broadly used and often considered as an alternative in comparable scenarios. 

Sadly tegola can not use minio, because minio does not use the bucket name as an subdomain, like AWS S3 does. 

Here is my docker-compose.yaml to demo the issue:
```
version: '3'
services:
  minio:
    image: minio/minio
    ports:
      - 9000:9000
    environment:
      MINIO_SECRET_KEY: ***
      MINIO_ACCESS_KEY: ***
    volumes:
      - ./minio/storage:/data
    entrypoint: sh
    command: -c 'mkdir -p /data/tegola && /usr/bin/minio server /data'
  tegola:
    image: gospatial/tegola
    ports:
      - 8080:8080
    volumes:
      - ./tegola/config:/config
    command: ["serve","--config","/config/config.toml"]
```

config.toml
```
[cache] 
type = "s3" 
bucket = "tegola"
endpoint = "http://minio:9000"
aws_access_key_id = "***"
aws_secret_access_key = "***"
```

With this configuration, tegola tries to connect to "http://tegola.minio:9000", but the correct path would be "http://minio:9000/tegola". This behavior can be changed by setting the aws-sdk-go S3Clients property **S3ForcePathStyle** to "true". 

With this issue I propose to add this parameter in an configurable way. I already have a working solution, which covers the parsing of the configuration parameter from config.toml and apply it in the same way as other configuration parameters. I would love to make PR. You can find my work here: [https://github.com/richardbischof/tegola/commit/36b05ba8e83c5b9cc2720d0971fcb5903431bc0f](https://github.com/richardbischof/tegola/commit/36b05ba8e83c5b9cc2720d0971fcb5903431bc0f)

I would love to have your feedback. 

Regards, Richard

ARolek:
[...]

PR #745, opened 2021-03-02
Allow forcing path style for s3 compatible API requests.

johngian:


gdey (approved):
LGTM

johngian:
I think that this patch fixes #719

ARolek:
@johngian excellent. Thanks for pointing out that issue.

PR #752, opened 2021-03-16
Initial prometheus

gdey:
Initial Prometheus integration


observability/prometheus/prometheus.go
```diff
@@ -0,0 +1,81 @@
...
+	prometheus.MustRegister(
+		inFlightGauge,
+		counter,
+		duration,
+		responseSize,
+	)
+}
+
+func Handler() (string, string, http.Handler) {
+	return http.MethodGet, "/metrics", promhttp.Handler()
```
ARolek, line 70: if we move "/metrics" to a var, we could make it configurable.

ARolek (commented):
I think we should split the HTTP observability into a separate PR. I think the main goal with this PR should be enabling Prometheus via the tegola config and registering the endpoint with the base Prometheus metrics.

README.md
```diff
@@ -118,6 +118,9 @@ ssl_key = "privkey.pem"     # ssl key for serving by https
   Access-Control-Allow-Origin = "*"
   Cache-Control = "no-cache, no-store, must-revalidate"
 
+[observe]
```
ARolek, line 121: this needs to be `observer` rather than `observe`

ARolek (approved):
LGTM!

PR #755, opened 2021-03-23
Add http end point observer

### LINKED ISSUE #754: Add Prometheus support for HTTP API routes and Viewer routes
**HTTP API routes**

The HTTP API routes will be prefixed with  `tegola_api_` and the following things will be observed

* flight_request is a gauge of the API requests currently being served
* requests_total is a counter of the total number of API requests that have been served, partitioned by code, and method
* duration_seconds is a histogram of the duration of the current API requests, partitioned by handler, code, and method. The handler label can further be partitioned by substituting various variables. Currently supported variables are `:map_name`, `:layer_name`, `:z`,`:x`,`:y`. Default set is `:map_name`, `:layer_name`, and `:z` 
* response_size_bytes is a histogram of response sized in bytes with the following buckets, 512 bytes, 1 KB, 10 KB, 1 MB, and 5+ MB.

**HTTP Viewer routes**

The HTTP API routes will be prefixed with  `tegola_viewer_` and the following things will be observed

* flight_request is a gauge of the Viewer requests currently being served
* requests_total is a counter of the total number of Viewer requests that have been served, partitioned by code, and method
* duration_seconds is a histogram of the duration of the current Viewer requests, partitioned by handler, code, and method. The handler is partitioned into `/` and `/*path`
* response_size_bytes is a histogram of response sized in bytes with the following buckets, 512 bytes, 1 KB, 10 KB, 1 MB, and 5+ MB.

The metric path initially will not be configurable and will be on the main host at `/metrics`

gdey:
[observability] Split out httpAPI and httpViewer instrumentation

Split out the Observability interface so that API route and dealt with separately from the
viewer routes.

Moved Null Observer to an internal package, treating nil as a Null Observer.
Prometheus observer can be configured to indicate which API vars should be filled in. Default variables
are :map_name, :layer_name and :z.

fixes #754

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -20,6 +20,12 @@ type Interface interface {
 	Purge(key *Key) error
 }
 
+// Wrapped Cache are for cache backend that wrap other cache backends
+// Original will return the first cache backend to be wrapped
+type Wrapped interface {
+	Original() Interface
+}
+
 // ParseKey will parse a string in the format /:map/:layer/:z/:x/:y into a Key struct. The :layer value is optional
 // ParseKey also supports other OS delimiters (i.e. Windows - "\")
 func ParseKey(str string) (*Key, error) {
--- a/observability/prometheus/prometheus.go
+++ b/observability/prometheus/prometheus.go
@@ -3,6 +3,8 @@ package prometheus
 import (
 	"net/http"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/go-spatial/tegola/observability"
 
 	"github.com/go-spatial/tegola/dict"
@@ -21,6 +23,9 @@ const (
 
 const (
 	Name = "prometheus"
+
+	httpAPI    = "tegola_api"
+	httpViewer = "tegola_viewer"
 )
 
 func init() {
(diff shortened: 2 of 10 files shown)


observability/prometheus/prometheus.go
```diff
@@ -31,19 +37,149 @@ func init() {
...
+			obs.responseSize = make(map[string]*prometheus.HistogramVec, 2)
+		}
+		// responseSize has no labels, making it a zero-dimensional
+		// ObserverVec.
+		obs.responseSize[name] = prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name: name + "_response_size_bytes",
+				Help: "A histogram of response sizes for requests.",
+				// TODO(gdey): Do we want to make this configurable? This is the buckets for tiles served up in bytes.
+				Buckets: []float64{512, float64(1 * kb), float64(10 * kb), float64(1 * mb), float64(5 * mb)},
```
ARolek, line 108: I don't think we need to make this configurable, but I do think we need a bucket at 500kb rather than 5mb. 500kb is usually a threshold to keep map tiles under. anything over 1mb can really be grouped together.

ARolek (commented):
This is looking really good. Just one comment inline.

cache/cache.go
```diff
@@ -20,6 +20,12 @@ type Interface interface {
 	Purge(key *Key) error
 }
 
+// Wrapped Cache are for cache backend that wrap other cache backends
+// Original will return the first cache backend to be wrapped
+type Wrapped interface {
```
ARolek, line 25: Maybe this question will be answered in some of the other PRs, but what's the plan for this interface?

ARolek: Got it. This could maybe be moved to the cache PR then. No biggie either way, I just didn't see it being used with this update which is why I asked about it.

gdey: I think I will leave here. But good point. The only reason I'm not removing it, is because I have to move it across three branches and rebases, and I'm running out of time.

ARolek: That's fair.

ARolek:
@gdey could you add a section to the http package README.md called something like "Prometheus Metrics" and explain the metrics that are available from this package

PR #765, opened 2021-03-30
Refactor build info

gdey:
Added a new internal package to hold build information.

Using the new build information package, expose a Observability Metric for the Build Information. So, that one can use that for doing analysis across versions.

See:  https://www.robustperception.io/exposing-the-software-version-to-prometheus


ARolek (approved):
LGTM

PR #771, opened 2021-04-06
Make sure to strip "/" for static files if URIPrefix ends with a "/"

### LINKED ISSUE #735: uri_prefix breaks with a trailing slash
When the uriPrefix has a trailing slash, the API server works but the static files become broken.

This breaks:

```
¶ cat config.toml 
[webserver]
uri_prefix = "/tegola/"
[…]
¶ docker-compose restart 
Restarting gnis_tegola_1 ... done
¶ curl localhost:8080/tegola/capabilities 
{"version":"v0.12.1", …
¶ curl localhost:8080/tegola/
404 page not found
```

But this works:

```
¶ cat config.toml 
[webserver]
uri_prefix = "/tegola"
[…]
¶ docker-compose restart 
Restarting gnis_tegola_1 ... done
¶ curl localhost:8080/tegola/capabilities 
{"version":"v0.12.1", …
¶ curl localhost:8080/tegola/
<!DOCTYPE html>…
```

gdey:
When serving up static files we need to strip any proxy prefixes we may add onto the URL.
We forgot to check to see if the prefix had a "/" at the end of it, and if it did made sure
to strip it out, as static files should always start with a "/". However, if we did not remove
the "/" at the end of the prefix that we then use to trim the url; we would end up trimming off
the starting "/". This fixes that.

Fixes #735

### DIFF
--- a/server/viewer.go
+++ b/server/viewer.go
@@ -29,8 +29,15 @@ type FilePathPrefixStripper struct {
 }
 
 func (fsps *FilePathPrefixStripper) Open(name string) (http.File, error) {
-	if URIPrefix != "/" {
-		name = strings.TrimPrefix(name, URIPrefix)
+	// Don't want to modify the Global.
+	// TODO(GDEY): Why is URIPrefix a global? Can it change after startup?
+	prefix := URIPrefix
+	// Strip any leading "/" from the URIPrefix.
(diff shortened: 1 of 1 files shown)


gdey:
Have not had a chance to test this yet.

Issue #758, opened 2021-03-23
Add metric end points for DataProviders


gdey:
For the Postgres provider here are sql_query break down:
Average query timing (breakdown by layer)

```
  Sub 100ms
  100ms -> 1sec
  1sec -> 5sec
  5sec > 20sec
  20sec >
```

Connection pool.
  Count

PR #780, opened 2021-05-25
[dataprovider] Added Observability to postgis data provider

gdey:
* Observability of the postgis pool
* MVTProvider has query time observability on a per map and zoom level
* Provider has query time observability on a per map, per layer and zoom level

### DIFF
--- a/atlas/map.go
+++ b/atlas/map.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/go-spatial/tegola/observability"
+
 	"github.com/golang/protobuf/proto"
 
 	"github.com/go-spatial/geom"
@@ -59,6 +61,8 @@ type Map struct {
 
 	mvtProviderName string
 	mvtProvider     provider.MVTTiler
+
+	observer observability.Interface
 }
 
 // HasMVTProvider indicates if map is a mvt provider based map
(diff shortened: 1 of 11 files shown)


gdey:
fixes #758

provider/postgis/postgis.go
```diff
@@ -22,14 +28,121 @@ import (
...
 	layers     map[string]Layer
 	srid       uint64
-	firstlayer string
+	firstLayer string
+
+	// registered keeps track if we have already registered these collectors
+	// as the Collectors function will be called for each map and layer, but
+	// we are going to assign those during runtime, instead of at registration
+	// time; so we will only return these collectors on the first call.
+	registered bool
```
ARolek, line 108: since this is collector specific, consider renaming to `collectorsRegistred` or similar.

ARolek (commented):
Looks good! I still need to run some tests on the mvt_provider implementation before I can approve this.

atlas/map.go
```diff
@@ -59,6 +61,9 @@ type Map struct {
 
 	mvtProviderName string
 	mvtProvider     provider.MVTTiler
+
+	obsLck   sync.Mutex
```
ear7h, line 65: Should this lock be used? I don't see it being used anywhere in the diffs

observability/observability.go
```diff
@@ -52,6 +56,11 @@ type Interface interface {
 	// Shutdown should shutdown any subsystems that Init setup
 	Shutdown()
 
+	// MustRegister will register any custom collectors. It will panic if there is an error
+	MustRegister(collectors ...Collector)
+
+	CollectorConfig(key string) map[string]interface{}
```
ear7h, line 62: Doc string for this

ear7h (commented):
I'm a little out of context, so I just have some minor comments

observability/observability.go
```diff
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"sort"
 
+	"github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
gdey: this is only here so that I can type Alias Collector as prometheus.Collector. Later we can change this out to an actual interface when we have more information on how observability will work with other providers.

atlas/map.go
gdey: Good catch

ear7h:
One thing I noticed, that I don't think is from this PR is a missing `nil` check here:
https://github.com/go-spatial/tegola/blob/5bfe906a356a1d0bd63230862f4bfcfa49103a6d/observability/prometheus/prometheus.go#L211
Because of the following line:
https://github.com/go-spatial/tegola/blob/5bfe906a356a1d0bd63230862f4bfcfa49103a6d/observability/prometheus/prometheus.go#L154

gdey:
> One thing I noticed, that I don't think is from this PR is a missing `nil` check here:
> 
> https://github.com/go-spatial/tegola/blob/5bfe906a356a1d0bd63230862f4bfcfa49103a6d/observability/prometheus/prometheus.go#L211
> 
> Because of the following line:
> 
> https://github.com/go-spatial/tegola/blob/5bfe906a356a1d0bd63230862f4bfcfa49103a6d/observability/prometheus/prometheus.go#L154

fixed.

ARolek:
@gdey I'm working on the local tests for this and realized the readme has not been updated with the new metrics. Please add those when you have a chance.

ARolek (approved):
Approved pending the documentation update!

PR #785, opened 2021-06-12
Utilize builtin go "embed" package to embed all static files needed by internal viewer UI

mapl:
dist folder not included in source


ARolek:
I'm a big fan of this PR, but we're in a peculiar situation where we need to keep supporting the old way of embedding assets. In an ideal situation, we could use go build flags to sniff the version of Go that's being used for building and use the `embed` package when the version is 1.16 or greater.

cc @gdey

ARolek (commented):
This looks good to me. No real comments. @gdey any feedback? You have more experience with `embed`.

server/viewer_embed.go
```diff
@@ -0,0 +1,19 @@
+// +build !noViewer,go1.16
```
gdey, line 1: Is there a reason why this is an OR and not an AND like embed.go.

`// +build !noViewer go1.16`?

server/viewer.go
```diff
@@ -1,4 +1,4 @@
-// +build !noViewer
+// +build !noViewer,!go1.16
```
gdey, line 1: shouldn't this be an AND and not an OR

`// +build !noViewer !go1.16`?

gdey (commented):
Need to look at the build failure. We may need to move the build of the assets around; seems to to be failing to find the `dist` dir according to the CI.

server/viewer_embed.go
mapl: It should be.

I had to look up the syntax for build constrains/tags. According to the docs and some help from the IDE, the rules are. 

A build constraint is **_evaluated as the OR of space-separated_** options. Each option evaluates as the **AND of its comma-separated** terms.

https://pkg.go.dev/cmd/go#hdr-Build_constraints

ui/embed.go
```diff
@@ -0,0 +1,23 @@
+// +build !noViewer go1.16
```
gdey, line 1: Then shouldn't this be `!noViewer,go1.16`?

server/viewer_embed.go
gdey: Okay, yeah I was confused. I had to read it several times.

ui/embed.go
mapl: Good catch. 
I have now completely omitted the +built constraint in embed.go
It doesn't seem to matter because "embed.go" is only referred in "viewer_embed.go" which is only considered when built with go 1.16+
It compiled fine with no tag set and with -tag NoViewer on Go1.15.14.

ARolek:
@mapl can you squash these commits when you have a chance? this is just about ready to be merged.

mapl:
I made some git mistakes and decided to open another PR for this.

PR #790, opened 2021-07-19
Refactor tegola_lambda CI to use Amazon Linux for building

### LINKED ISSUE #789: Aws lambda Execution failed - /lib64/libc.so.6: version `GLIBC_2.28' not found
Trying to run Tegola with AWS lambda.  
i got this error when testing : 
`
START RequestId: 1340f6e4-ceee-401b-950b-cac0880a4f95 Version: $LATEST
/var/task/tegola_lambda: /lib64/libc.so.6: version 'GLIBC_2.28' not found (required by /var/task/tegola_lambda)
END RequestId: 1340f6e4-ceee-401b-950b-cac0880a4f95
REPORT RequestId: 1340f6e4-ceee-401b-950b-cac0880a4f95	Duration: 35.30 ms	Billed Duration: 36 ms	Memory Size: 512 MB	Max Memory Used: 6 MB	
RequestId: 1340f6e4-ceee-401b-950b-cac0880a4f95 Process exited before completing request
`

i tried the version from github and a compiled version with go 1.3 and go 1.6.  
also tried the option "CGO_ENABLED=0" on the compilation without success

ARolek:
AWS Lambda uses Amazon Linux for Lambda executions. The previous
build pipeline was using Ubuntu 18.04 to build tegola_lambda
which had GLIB location parity with Amazon Linux. After upgrading
to Ubuntu 20.04 the GLIB references between Ubuntu and Amazon Linux
are no longer in parity so tegola_lambda fails to find GLIB during
Lambda execution.

This commit accomplished the following:

* build tegola_lambda using Amazon Linux rather than Ubuntu
* move from using ubuntu-latest to ubuntu-20.04 in GH actions.

closes #789

### DIFF
--- a/.github/actions/amazon-linux-build-action/Dockerfile
+++ b/.github/actions/amazon-linux-build-action/Dockerfile
@@ -1,0 +1,22 @@
+# Amazon Linux is used to build tegola_linux so the CGO requirements are linked correctly
+FROM amazonlinux:latest
+
+# install build deps
+RUN yum install -y tar gzip gcc
+
+# install Go
+ENV GOLANG_VERSION 1.16.6
+ENV GOLANG_VERSION_SHA256 be333ef18b3016e9d7cb7b1ff1fdb0cac800ca0be4cf2290fe613b3d069dfe0d
+
+RUN curl -o golang.tar.gz https://dl.google.com/go/go$GOLANG_VERSION.linux-amd64.tar.gz \
+	&& echo "$GOLANG_VERSION_SHA256 golang.tar.gz" | sha256sum --strict --check \
+	&& tar -C /usr/local -xzf golang.tar.gz \
+	&& rm golang.tar.gz
+
+ENV PATH /usr/local/go/bin:$PATH
+
+# entrypoint.sh holds the build instructions for tegola_lambda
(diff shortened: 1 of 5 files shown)


.github/actions/amazon-linux-build-action/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Amazon Linux is used to build tegola_linux so the CGO requirements are linked correctly
+FROM amazonlinux:latest
```
gdey, line 2: Should we be using the latest here? That is what got us in trouble with ubuntu?

gdey (approved):
LGTM; only one minor comment about using the latest tag for amazonlinux, instead of specifying a version. This way we aren't caught by surprise again.

.github/actions/amazon-linux-build-action/Dockerfile
ARolek: I debated this, precisely because of the `ubuntu-latest` issue. My thinking is that the point of this container is to build tegola_lambda using the most recent Amazon Linux to avoid any potential linking issues. I don't expect Amazon Linux to have breaking changes in the near future, but a version bump from say Amazon Linux 2 to Amazon Linux 3 *should* work automatically.

PR #795, opened 2021-08-03
Feature: adds configurable query parameters to tile endpoints

dechristopher:
Hi all. I recently stumbled upon a use-case for having query parameters present in URLs. Being able to configure Tegola's queries on the fly with optional URL query parameters for filtering down data can simplify configuration for large datasets that would otherwise have many layers defined manually.

Defining a query parameter is as simple as this:
```toml
[[maps.params]]
name = "param"
token = "!PARAM!"
default = "7"
```

This will configure a parameter named `param` with the SQL replacement token `!PARAM!`. The `default` key is optional and allows you to specify a default value for the parameter if none is provided. Without a default specified, the replacer will default to substituting an empty string.

The endpoint then becomes queryable at:
```
http://server:8080/maps/example/{z}/{x}/{y}.pbf?param=123
```

Currently, to prevent SQL injection, parameter values are limited to values matching the following regular expression. This is to limit control characters from being injected via query parameters:
```regex
[a-zA-Z0-9,._-]+
```

This obviously needs more work and I am open to feedback regarding security. I've also thought of limiting the scope of query parameters to integers only since most of the use cases I've found for this feature involve numeric ID matching in queries.

Any and all feedback is welcome. I hope others can find value in this contribution.

### DIFF
--- a/provider/postgis/util.go
+++ b/provider/postgis/util.go
@@ -174,6 +174,46 @@ func replaceTokens(sql string, lyr *Layer, tile provider.Tile, withBuffer bool) 
 	return tokenReplacer.Replace(uppercaseTokenSQL), nil
 }
 
+// replaceParams substitutes configured query parameter tokens for their values
+// within the provided SQL string
+func replaceParams(ctx context.Context, sql string) (string, error) {
+	if ctx.Value("params") == nil {
+		return sql, nil
+	}
+
+	var params map[string]string
+	params = ctx.Value("params").(map[string]string)
+
+	replacerVals := make([]string, 0)
+
+	for token, val := range params {
+		if val != "" && !paramValRe.Match([]byte(val)) {
+			return "", fmt.Errorf("Param value for token %s contains illegal characters: %s ",
+				token, val)
+		}
+
+		replacerVals = append(replacerVals, token, val)
+	}
+
+	// generate replacer from token/value pairs
+	replacer := strings.NewReplacer(replacerVals...)
+
+	uppercaseTokenSQL := uppercaseTokens(sql)
+
+	return replacer.Replace(uppercaseTokenSQL), nil
+}
+
+// stripParams will remove all parameter tokens from the query
+func stripParams(sql string) string {
+	return tokenRe.ReplaceAllStringFunc(sql, func(s string) string {
+		return ""
+	})
+}
+
+// paramValRe restricts parameter values to a limited character set to prevent
+// SQL injection attacks. TODO This is very primitive and needs more thought.
+var paramValRe = regexp.MustCompile("[a-zA-Z0-9,._-]+")
+
 var tokenRe = regexp.MustCompile("![a-zA-Z0-9_-]+!")
 
 //	uppercaseTokens converts all !tokens! to uppercase !TOKENS!. Tokens can
--- a/provider/postgis/util.go
+++ b/provider/postgis/util.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-spatial/geom"
 	"github.com/go-spatial/tegola"
 	"github.com/go-spatial/tegola/basic"
+	"github.com/go-spatial/tegola/config"
 	"github.com/go-spatial/tegola/provider"
 	"github.com/jackc/pgx"
 	"github.com/jackc/pgx/pgtype"
@@ -97,20 +98,6 @@ func genSQL(l *Layer, pool *connectionPoolCollector, tblname string, flds []stri
 	return fmt.Sprintf(sqlTmpl, selectClause, tblname, l.geomField), nil
 }
 
-const (
-	bboxToken             = "!BBOX!"
-	zoomToken             = "!ZOOM!"
-	xToken                = "!X!"
-	yToken                = "!Y!"
-	zToken                = "!Z!"
-	scaleDenominatorToken = "!SCALE_DENOMINATOR!"
-	pixelWidthToken       = "!PIXEL_WIDTH!"
-	pixelHeightToken      = "!PIXEL_HEIGHT!"
-	idFieldToken          = "!ID_FIELD!"
-	geomFieldToken        = "!GEOM_FIELD!"
-	geomTypeToken         = "!GEOM_TYPE!"
-)
-
 // replaceTokens replaces tokens in the provided SQL string
 //
 // !BBOX! - the bounding box of the tile
@@ -169,17 +156,17 @@ func replaceTokens(sql string, lyr *Layer, tile provider.Tile, withBuffer bool) 
 	// replace query string tokens
 	z, x, y := tile.ZXY()
 	tokenReplacer := strings.NewReplacer(
-		bboxToken, bbox,
-		zoomToken, strconv.FormatUint(uint64(z), 10),
-		zToken, strconv.FormatUint(uint64(z), 10),
-		xToken, strconv.FormatUint(uint64(x), 10),
-		yToken, strconv.FormatUint(uint64(y), 10),
-		idFieldToken, lyr.IDFieldName(),
-		geomFieldToken, lyr.GeomFieldName(),
-		geomTypeToken, geoType,
-		scaleDenominatorToken, strconv.FormatFloat(scaleDenominator, 'f', -1, 64),
-		pixelWidthToken, strconv.FormatFloat(pixelWidth, 'f', -1, 64),
-		pixelHeightToken, strconv.FormatFloat(pixelHeight, 'f', -1, 64),
+		config.BboxToken, bbox,
+		config.ZoomToken, strconv.FormatUint(uint64(z), 10),
+		config.ZToken, strconv.FormatUint(uint64(z), 10),
+		config.XToken, strconv.FormatUint(uint64(x), 10),
+		config.YToken, strconv.FormatUint(uint64(y), 10),
+		config.ScaleDenominatorToken, strconv.FormatFloat(scaleDenominator, 'f', -1, 64),
+		config.PixelWidthToken, strconv.FormatFloat(pixelWidth, 'f', -1, 64),
+		config.PixelHeightToken, strconv.FormatFloat(pixelHeight, 'f', -1, 64),
+		config.IdFieldToken, lyr.IDFieldName(),
+		config.GeomFieldToken, lyr.GeomFieldName(),
+		config.GeomTypeToken, geoType,
 	)
 
 	uppercaseTokenSQL := uppercaseTokens(sql)
--- a/server/handle_map_layer_zxy.go
+++ b/server/handle_map_layer_zxy.go
@@ -156,8 +156,29 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		m = m.AddDebugLayers()
 	}
 
-	encodeCtx := context.WithValue(r.Context(), observability.ObserveVarMapName, m.Name)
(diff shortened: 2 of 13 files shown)


ARolek:
@dechristopher awesome! This feature has been [requested before](https://github.com/go-spatial/tegola/issues/610) so it's great you tackled it! I should be able to get a review in the next week. 

cc @gdey @ear7h

dechristopher:
> @dechristopher awesome! This feature has been [requested before](https://github.com/go-spatial/tegola/issues/610) so it's great you tackled it! I should be able to get a review in the next week.
> 
> cc @gdey @ear7h

Any chance to review yet?

gdey:
I'll look at this Friday. @dechristopher  do you know why the test is failing? Have not dug into the logs. But a quick scan is complaining about bbox. I know you are dynamically changing that in the SQL.

dechristopher:
> I'll look at this Friday. @dechristopher do you know why the test is failing? Have not dug into the logs. But a quick scan is complaining about bbox. I know you are dynamically changing that in the SQL.

Unsure why the tests are failing on the runner. They run cleanly on my machine. I'll investigate and see if I can come up with a fix.

ear7h:
This is a really cool feature! Thanks for getting it started!

WRT to input validation I think the best approach is to use the "placeholder" mechanism offered by pgx (it's described briefly in  [`pgx.Query`](https://pkg.go.dev/github.com/jackc/pgx/v4#Conn.Query)). I think the biggest difficulty with this is getting the numbering right, and keeping track of it. I've come up with a function that could do that, but I'm not sure how it would fit with that part of the code base:

https://play.golang.org/p/Dk6SWSf2f_N

This is also missing a few things like case insensitivity and special handling of the builtin tokens. The later could be implemented by writing to the `queryArgs` map after writing the user provided args (so any colliding user provided args get overwritten).

provider/postgis/util.go
```diff
@@ -169,24 +156,64 @@ func replaceTokens(sql string, lyr *Layer, tile provider.Tile, withBuffer bool)
...
 
 	uppercaseTokenSQL := uppercaseTokens(sql)
 
 	return tokenReplacer.Replace(uppercaseTokenSQL), nil
 }
 
+// replaceParams substitutes configured query parameter tokens for their values
+// within the provided SQL string
+func replaceParams(ctx context.Context, sql string) (string, error) {
+	if ctx.Value("params") == nil {
```
ARolek, line 180: maybe make the value "params" a const since you're referencing it multiple times.

provider/postgis/util.go
```diff
@@ -169,24 +156,64 @@ func replaceTokens(sql string, lyr *Layer, tile provider.Tile, withBuffer bool)
...
+		return sql, nil
+	}
+
+	var params map[string]string
+	params = ctx.Value("params").(map[string]string)
```
ARolek, line 185: add an `!ok` check when doing an assertion.

provider/postgis/util.go
```diff
@@ -169,24 +156,64 @@ func replaceTokens(sql string, lyr *Layer, tile provider.Tile, withBuffer bool)
...
+
+	replacerVals := make([]string, 0)
+
+	for token, val := range params {
+		if val != "" && !paramValRe.Match([]byte(val)) {
+			return "", fmt.Errorf("Param value for token %s contains illegal characters: %s ",
```
ARolek, line 191: nit, you're changing the first letter of the error text to lowercase in other parts of the codebase, but using uppercase here. 😜

server/handle_map_layer_zxy.go
```diff
@@ -156,8 +156,36 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	encodeCtx := context.WithValue(r.Context(), observability.ObserveVarMapName, m.Name)
-	pbyte, err := m.Encode(encodeCtx, tile)
+	ctx := context.WithValue(r.Context(), observability.ObserveVarMapName, m.Name)
+	var params map[string]string
+
+	// check for query parameters and populate param map with their values
+	if req.Atlas.HasParams(req.mapName) {
+		params = make(map[string]string)
+		err = r.ParseForm()
```
ARolek, line 165: Any reason why you went with r.ParseForm() instead of using [r.URL.Query()](https://pkg.go.dev/net/url#URL.Query)?

provider/postgis/util.go
ARolek: After reading the commit, I'm not sure if passing the params around via the context is the best approach. I know this allows us to avoid updating the provider interfaces, but I'm generally not a fan of stashing values into context.

cc: @gdey @ear7h

ARolek (changes requested):
This is a great addition, so thank you for taking the time to submit this PR as well as clean up some parts of the codebase! I added a few comments inline, most are fairly minor. The main issue I would like to discuss is the use of context for passing around the `params`. I think it might be better to update the feature provider to support a new parameter for filters rather than use context.

ARolek:
Another consideration with this PR is going to be the tile cache. If the query params can produce different features based on the query strings, we're going to need to revisit the tile cache. Alternatively, we can say if filter params are used, then no caching is allowed right now.

ARolek:
@dechristopher and @kszafran apologies for the slow response. I missed the comments you two dropped in here. I'm interested in bringing this feature to life as well, but I posted some concerns with the current implementation. It might be easier to work on the design in real-time via slack. We can keep the discussion here as well, but I don't think we should be using `context.Context` to pass around the filters. I think we should enhance the `Tilter` interface:

```go
// Tiler is a Layers that allows one to encode features in that layer
type Tiler interface {
	Layerer

	// TileFeature will stream decoded features to the callback function fn
	// if fn returns ErrCanceled, the TileFeatures method should stop processing
	TileFeatures(ctx context.Context, layer string, t Tile, fn func(f *Feature) error) error
}
```

We can add another parameter to `TileFeatures` for filters. I'm not sure what the design for Filters should be yet, but I think this thread already has some ideas brewing. Ideally we can figure out a design that allows us to filter not only SQL providers but other future providers too (i.e. GeoJSON). We don't need to implement the filters now, but I would like to consider them.

bemyak:
Hi!

I'm starting working to push this feature forward. The main issue currently is the SQL injection vulnerability. We definitely need to use prepared statements, i.e., replace all user-defined parameters with `$1`, `$2` and so on, and pass the values as arguments.

Unfortunately, this means that we need to have type information available for parameters.

So far the idea is to have parameters configured like this:
```toml
[[maps.params]]
name = "param2"
token = "!PARAM2!"
# only very simply types are suppored: `int`, `string`, `float`, `bool`
type = "int"
# `?` will be replaced with the actual value
# if `sql` is not specified, it is considered to be `?`
sql = "AND ANSWER = ?"
# if the parameter value is missing from the HTTP query, this value will be used
# `default_value` is an optional parameter. If it is not defined, the parameter is considered as required and cliet will get 400 error
default_value = "42"
# instead of `default_value`, `default_sql` can be defined, e.g. to omit param query entirely:
#default_sql = ""
# defining both `default_value` and `default_sql` is forbidden
```

I will be really glad to hear any feedback or thoughts on this.

kszafran:
I'm working with @bemyak on preparing the spec for this. Another approach that I would propose is like this:
```toml
[[maps.params]]
name = "param2"
token = "!PARAM2!"
type = "int"
# if `sql` is not specified, it is `?` by default, so just the param value and not a whole expression
sql = "AND answer = ?"
# if the parameter is not specified by the caller, this value will be used instead
# if "default" is not defined, the `!PARAM2!` will be replaced with an empty string instead of "sql"
default = "42"
# "required" could be used in cases where we don't want to specify a "default" value, but want to make the parameter mandatory
required = "true"
```

The main difference compared to @bemyak proposal is how we behave when there's no `default` and a caller does not specify the parameter. There are a few different cases that we like to handle:
1. parameter is required
2. parameter is optional and we use a default value if it's not specified
3. parameter is optional and we skip the whole condition if it's not specified

Those can be achieved with both designs. I just think that having only `default` plus replacing the token with an empty string would be simpler than having both `default_value` and default_sql`. Here's how you could achieve the three use cases:

1:
```toml
sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) ... AND answer = !ANSWER!"
...
[[maps.params]]
name = "answer"
token = "!ANSWER!"
type = "int"
required = "true"
```

2:
```toml
sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) ... AND answer = !ANSWER!"
...
[[maps.params]]
name = "answer"
token = "!ANSWER!"
type = "int"
default = "42"
```

3:
```toml
sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) ... !ANSWER!"
...
[[maps.params]]
name = "answer"
token = "!ANSWER!"
type = "int"
sql = "AND answer = ?"
```

ARolek:
@bemyak and @kszafran apologies for the slow response and thanks for the effort on this issue. 

### Additional context to this discussion: 

* [The original issue around filtering a layer dynamically](https://github.com/go-spatial/tegola/issues/610)
* [A PR that was submitted which has review comments](https://github.com/go-spatial/tegola/pull/795): I recall the main issue I had with this PR was the abuse of context.Context. I think we need to figure out a more defined way to pass this information around rather than stuffing it into context. I think enhancing the TileProvider interface is the way to go.

### Additional considerations:

* When dynamic tile filtering is implemented, we need to think about how the tile cache is supposed to work. Currently the tile cache is either enabled or disabled. This is not configured on a per map basis. It seems to me that enabling tile filtering will require the tile cache to be off. If this is the case, I think we need to investigate if we want to support cache configs on a per map basis so we can allow for some maps to leverage the cache but others to have dynamic filtering.

* SQL injections are a very real problem with this feature. I'm trying to figure out where the responsibility should lie for sanitizing the input. Tegola could be responsible for this as you're proposing. Alternatively we could use Postgres functions and pass on the responsibility to the database function. This might be more prone to error though as it's pushing the security responsibility to the implementer, but it would simplify the implementation in tegola. We might just be able to get away with a "query string whitelist" in the config that would then look for query strings and then turn them into tokens. Here's an example of calling a database function from tegola:

```sql
SELECT * FROM my_tile_func(!ZOOM!, !BBOX!, !FILTER_1!, !FILTER_2!)
```

I have used this pattern in tegola to push complex query logic to the database and it's quite nice. 

### Let's chat in real time:

This feature is one I would like to get right. I see the value, and I'm happy to support the effort. I do think it deserves some strong vetting though as the feature can be prone to abuse. Maybe we should try to chat through some of this in real time. You can find us in a couple of Slack channels:

* https://slack.go-spatial.org/
* The #go-spatial channel in Gopher Slack: https://gophers.slack.com

kszafran:
@ARolek Thanks for the comments! @bemyak is actively working on a PR. It's almost ready and we should soon be able to create a PR to the upstream. If you want to take a look already, check out https://github.com/SharperShape/tegola/pull/6.

How about this - once we're ready, we'll create a PR to the upstream, and then we could have a call to discuss it (e.g. on Google Meet)? Slack is fine, too. Our implementation solves SQL injection by using query arguments instead of putting parameter values into the query string itself. For now, caching is disabled for parameterized maps, but I imagine that could be improved in the future.

ARolek:
> How about this - once we're ready, we'll create a PR to the upstream, and then we could have a call to discuss it (e.g. on Google Meet)?

Love it. Once we get there we can see who can all join the discussion. Thanks for working on this contribution.

gdey:
#867 supersedes this PR.

Issue #798, opened 2021-08-13
Link in attribution


underspica:
Is there a way to display link in attribution field like [Mapbox](https://docs.mapbox.com/mapbox-gl-js/example/simple-map/) or [MapTiler](https://openmaptiles.github.io/osm-bright-gl-style/)?
I think this line escape all html tag: https://github.com/go-spatial/tegola/blob/v0.14.x/cmd/internal/register/maps.go#L14

ARolek:
@underspica I think you're right. We should probably have some form of sanitization but I think escaping HTML is the wrong thing to do there. Would you be interested in sending in a PR? We're really close to cutting the v0.14.0 release so you could probably sneak it in.

underspica:
Here is the PR #806
and some test cases https://play.golang.org/p/QL8w1w6xd54

PR #806, opened 2021-08-20
Allow link in attribution

underspica:
Issue #798
Allow link in attribution by unescape anchor tag.

### DIFF
--- a/cmd/internal/register/maps.go
+++ b/cmd/internal/register/maps.go
@@ -2,6 +2,8 @@ package register
 
 import (
 	"html"
+	"regexp"
+	"strings"
 
 	"github.com/go-spatial/geom"
 	"github.com/go-spatial/tegola/atlas"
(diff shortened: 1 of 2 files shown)


cmd/internal/register/maps.go
```diff
@@ -11,7 +13,14 @@ import (
 
 func webMercatorMapFromConfigMap(cfg config.Map) (newMap atlas.Map) {
 	newMap = atlas.NewWebMercatorMap(string(cfg.Name))
-	newMap.Attribution = html.EscapeString(string(cfg.Attribution))
+
+	attribution := html.EscapeString(string(cfg.Attribution))
+	re := regexp.MustCompile(`&lt;(a\s(.+?)|/a)&gt;`)
```
ARolek, line 18: Please move the regexp.MustCompile outside the function call.

cmd/internal/register/maps.go
```diff
@@ -11,7 +13,14 @@ import (
...
+	tags := re.FindAllString(attribution, -1)
+	for _, tag := range tags {
+		attribution = strings.Replace(attribution, tag, html.UnescapeString(tag), 1)
+	}
```
ARolek, line 22: You could move this to a separate function and maybe call it `sanitizeAttribution`. That would allow you to add your test cases to the codebase which would be ideal.

ARolek (changes requested):
A couple of minor comments added inline. Please add tests cases. Thanks for the contribution!

underspica:
Please review the new commits

ARolek:
@underspica this looks excellent! Thanks for making the adjustments. The last request is to squash everything into a single commit then I will get this merged in. Thank you!

PR #808, opened 2021-09-10
gpkg: improve error messages in registration

### LINKED ISSUE #805: Improve error message for gpkg register
I just ran into the panic from #761 because I misspelled something in my config. 

My suggestion would be to add a check to this block
https://github.com/go-spatial/tegola/blob/a64417e76597128814ac1cf8154b7e612460700c/provider/gpkg/gpkg_register.go#L312-L323

along the lines of:
```go
_, ok := geomTableDetails[tablename]
if !ok {
    return nil, errors.New("table " + tablename + " does not exist")
}
```

ear7h:
closes #805

I also added support for floating point tag columns (I came across this in the natural earth dataset)


ARolek (approved):
LGTM!

PR #820, opened 2021-12-28
chore: update pgx to v4

iwpnd:
Hey, as discussed I took a look at the pgx upgrade you wanted to do. It involved some refactoring as you expected.

Some things that took me a while were:

- `pgx`, `pgtype` and `pgproto3` are now separate packages
- hstore as an extension does not have static OID, so the OID has to be fetched after connect and then registered as a datatype. It's the first time I encounter something like this, so I'd love to get your opinion here.
- the `pgxpool` is instantiated with a connection string. For now I'm creating the string from the config parameters to allow backward compatibility. This now lends itself to allow a connection string in the `config.toml`.
- it might be worth rethinking the `ConfigTLS()` as most of the options are gone now, and fallback connections are handled differently


ARolek:
Well, this is incredible! I will get a review done this week. Thank you for the contribution!!!

gdey (approved):
Did a quick visual scan through the code. Nothing major jumps out at me.

Thank you for this!

LGTM

provider/postgis/postgis.go
```diff
@@ -659,7 +684,7 @@ func (p Provider) TileFeatures(ctx context.Context, layer string, tile provider.
 	}
 
 	now := time.Now()
-	rows, err := p.pool.Query(sql)
+	rows, err := p.pool.Query(context.Background(), sql)
```
ARolek, line 687: You have access to the actual context here, so we should use that instead of `context.Background()`

provider/postgis/postgis.go
```diff
@@ -814,7 +839,7 @@ func (p Provider) MVTForLayers(ctx context.Context, tile provider.Tile, layers [
 	}
 	{
 		now := time.Now()
-		err = p.pool.QueryRow(fsql).Scan(&data)
+		err = p.pool.QueryRow(context.Background(), fsql).Scan(&data)
```
ARolek, line 842: You have access to the actual context here, so we should use that instead of context.Background()

provider/postgis/util_internal_test.go
```diff
@@ -123,23 +124,25 @@ func TestDecipherFields(t *testing.T) {
...
-		Password: os.Getenv("PGPASSWORD"),
-	}
+	host := os.Getenv("PGHOST")
+	port := 5432
+	db := os.Getenv("PGDATABASE")
+	user := os.Getenv("PGUSER")
+	password := os.Getenv("PGPASSWORD")
+
+	cs := fmt.Sprintf("postgres://%v:%v@%v:%v/%v", user, password, host, port, db)
+	dbconfig, _ := BuildDBConfig(cs)
```
ARolek, line 134: any specific reason to ignore the error here?

ARolek (changes requested):
This looks really good! I have done a pass on the code and left some minor comments inline. I will work on some local QA next. 

👏

ARolek:
@iwpnd can you rebase the `v0.15.x` branch onto this one? I updated our CI so it would build some test binaries on PRs. That way if someone wants to help with QA they don't need to build from source.

iwpnd:
How do you do local QA @ARolek? Maybe something I can help you with?

ARolek:
@iwpnd apologies for the slow response here. I had to vacate my family from the [Marshall Fire](https://en.wikipedia.org/wiki/2021%E2%80%932022_Boulder_County_fires#Marshall_Fire). Our home survived but over 1000 homes are gone. Truly a sad situation.

For local QA I'm planning on spinning up Postgres 13 and Postgres 14 with some datasets to make sure everything works as expected. Did you do any of this type of testing locally?

iwpnd:
> @iwpnd apologies for the slow response here. I had to vacate my family from the [Marshall Fire](https://en.wikipedia.org/wiki/2021%E2%80%932022_Boulder_County_fires#Marshall_Fire). Our home survived but over 1000 homes are gone. Truly a sad situation.

Don't apologize for that. Glad to hear you and your family are okay. 

> For local QA I'm planning on spinning up Postgres 13 and Postgres 14 with some datasets to make sure everything works as expected. Did you do any of this type of testing locally?

Spun up a pg13/PostGIS3.1 instance with some of our data and it worked. Wouldn't call it testing yet though. No TLS connection or anything involved. When I'm back from vacation I thought about forking an image with the change and run it on our staging cluster with some more live data.

ARolek:
I have successfully tested this PR against Postgres 14 using the Natural Earth dataset. Looking good! I still need to test the TLS support, but no issues encountered thus far.

iwpnd:
I'll be forking the image today and will try to connect to our pg13.4/3.1 cluster using TLS. I'll keep you posted here.

ARolek:
@iwpnd I just saw your update and doing some additional testing I'm hitting the same issue. I found the fix. We need to change our error statements from:

```go
return fmt.Errorf("error replacing layer tokens for layer (%v) SQL (%v): %v", layer, sql, err)
```

To

```go
return fmt.Errorf("error replacing layer tokens for layer (%v) SQL (%v): %w", layer, sql, err)
```

Note the change of `%v` -> `%w`. This wraps the error check in `atlas/map.go` fails to identify the `context.Canceled` error:

```go
			if err != nil {
				switch {
				case errors.Is(err, context.Canceled):
					// Do nothing if we were cancelled.

				default:
					z, x, y := tile.ZXY()
					// TODO (arolek): should we return an error to the response or just log the error?
					// we can't just write to the response as the waitgroup is going to write to the response as well
					log.Printf("err fetching tile (z: %v, x: %v, y: %v) features: %v", z, x, y, err)
				}
				return
			}
```

Can you please update the errors in the `postgis` package to use wrapping?

iwpnd:
While this is not the error that is raised on my end, I see what you mean and will address it.

My error occurs in `atlas/handle_map_layer_zxy.go` and following your example this:

```go
	if err != nil {
		switch err {
		case context.Canceled:
			// TODO: add debug logs
			return
		default:
			errMsg := fmt.Sprintf("error marshalling tile: %v", err)
			log.Error(errMsg)
			http.Error(w, errMsg, http.StatusInternalServerError)
			return
		}
	}
```

should actually be:

```go
	if err != nil {
		switch {
		case errors.Is(err, context.Canceled):
			// TODO: add debug logs
			return
		default:
			errMsg := fmt.Sprintf("error marshalling tile: %v", err)
			log.Error(errMsg)
			http.Error(w, errMsg, http.StatusInternalServerError)
			return
		}
	}
```

right? This solved my error at least.

ARolek:
> Executing a query by not passing the ctx gets rid of the error, but I do not yet have the necessary understanding as to why and its implication.

Can you log the error type so we can see what pgx is returning?

```go
log.Printf("error is of type %T, err: %v", err, err)
```

We might just need to add an additional check if pgx is not wrapping the context.Canceled error correctly.

iwpnd:
Hey @ARolek, sorry for the late reply, I'm back on the grind after vacation.

```
2022-01-11 09:22:13 [ERROR] handle_map_layer_zxy.go:169: error marshalling tile: failed to connect to `host=localhost user=samson database=geometries`: dial error (dial tcp 127.0.0.1:5433: operation was canceled) (*pgconn.connectError)
```

gdey:
> Hey @ARolek, sorry for the late reply, I'm back on the grind after vacation.
> 
> ```
> 2022-01-11 09:22:13 [ERROR] handle_map_layer_zxy.go:169: error marshalling tile: failed to connect to `host=localhost user=samson database=geometries`: dial error (dial tcp 127.0.0.1:5433: operation was canceled) (*pgconn.connectError)
> ```

Looking at this you may need to unwrap the error a bit more. It looks like the `connectError` properly has an [Unwrap method](https://github.com/jackc/pgconn/blob/a1852214fe10eeeb5b7f8634a8dd1e928884f2cf/errors.go#L60), so one of the errors returned by the error being cased [here](https://github.com/jackc/pgconn/blob/a1852214fe10eeeb5b7f8634a8dd1e928884f2cf/pgconn.go#L239) is not wrapping the context.Cancelled error correctly. I cannot dig further as `DialFunc` is dynamic to the configuration.

ARolek:
I also found this related issue: https://github.com/jackc/pgx/issues/933. I think what @gdey is saying might be correct, we're going to need to unwrap the errors from the `pgx` client and check if they're context canceled errors to catch this. Seems not ideal. If we don't handle it, I think we're going to see a lot of these context canceled errors in the logs.

ARolek:
I made a local update to wrap the errors in the messages with `%w` instead of `%v`. I'm still seeing a bunch of error logs though:

```
map.go:330: err fetching tile (z: 7, x: 23, y: 47) features: error running layer (ne_10m_land) SQL (SELECT ST_AsBinary(wkb_geometry) AS geometry, ogc_fid, featurecla, min_zoom FROM ne_10m_land WHERE wkb_geometry && ST_MakeEnvelope(-1.2841420750122068e+07,5.00448511519043e+06,-1.2518550742690431e+07,5.32735512262207e+06,3857)): failed to connect to `host=<redacted> user=<redacted> database=natural_earth`: dial error (dial tcp 192.168.5.5:5432: operation was canceled)
```

"Operation was canceled" does not seem to be a `context.Canceled` error though. I'm also seeing:

```
2022/01/12 15:20:02 map.go:330: err fetching tile (z: 9, x: 86, y: 189) features: error running layer (ne_10m_admin_1_label_points) SQL (SELECT ST_AsBinary(wkb_geometry) AS geometry, ogc_fid, name, scalerank FROM ne_10m_admin_1_label_points WHERE wkb_geometry && ST_MakeEnvelope(-1.3307380874483643e+07,5.164697126453858e+06,-1.3226663372625733e+07,5.245414628311767e+06,3857)): failed to connect to `host=<redacted> user=<redacted> database=natural_earth`: failed SASL auth (write tcp 192.168.2.92:60731-><redacted>:5432: i/o timeout)
```

That does not happen on startup, but I believe as the pool is allocating more connections this error is showing up for each new connection being added to the pool. I'm still investigating but want to share my notes.

iwpnd:
https://github.com/jackc/pgx/issues/1109

Apparently we need to check `rows.Err()` after `rows.Next()`, maybe a pointer. Because:

> Not all errors can be detected when Query returns. You must read the results of the query and then check rows.Err() after that.

Regarding the TLS connection. There's been a major change in v4 as to how fallback connections and TLS is handled. As I noted in my initial comment, I only skip the verification.

ARolek:
> Apparently we need to check rows.Err() after rows.Next(), maybe a pointer. Because:

I have reviewed the code and we are doing this already. It's usually the last statement in the method:

```go
return rows.Err()
```

iwpnd:
I took a look at your changes and they were not sufficient to quiet the errors. Both the connection and context canceled was spammed continuesly:

```
2022-01-14 08:24:53 [ERROR] handle_map_layer_zxy.go:168: error marshalling tile: timeout: context canceled
2022-01-14 08:24:53 [ERROR] handle_map_layer_zxy.go:168: error marshalling tile: failed to connect to `host=localhost user=samson database=geometries`: dial error (dial tcp 127.0.0.1:5433: operation was canceled)
2022-01-14 08:24:57 [ERROR] handle_map_layer_zxy.go:168: error marshalling tile: timeout: context canceled
``` 

I took your advice and did two things. First I made sure that i handle the error in `dbconfig.AfterConnect` properly. Next up, I took a look at the switch statement in `handle_map_layer_xyz` where the error was originating and use the same logic that you did in `atlas/map.go` which quieted the errors entirely. Can you please take a look at it too?

Why exactly is the context canceled here anyways, and what does it entail?

ARolek:
> Why exactly is the context canceled here anyways, and what does it entail?

Good question. So context Canceleing is really important in this codebase as it accounts for the situation when users pan the map before a tile has been fully fetched, encoded and returned to the client. Without proper context handling, we're processing requests that are no longer needed.

server/handle_map_layer_zxy.go
```diff
@@ -158,10 +159,15 @@ func (req HandleMapLayerZXY) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		switch err {
-		case context.Canceled:
+		switch {
+		case errors.Is(err, context.Canceled):
 			// TODO: add debug logs
+			// do nothing
+			return
+		case strings.Contains(err.Error(), "operation was canceled"):
+			// do nothing
```
gdey, line 170: We are trying to get away from this if possible.

ARolek: @gdey yeah, I know, but we can't in this instance. I left a code comment around this situation in `map.go`:

```go
// the underlying net.Dial function is not properly reporting
// context.Cancel errors. Becuase of this, a string check on the error is performed.
// there's an open issue for this and it appears it will be fixed eventually
// but for now we have this check to avoid unnecessary logs
// https://github.com/golang/go/issues/36208
case strings.Contains(err.Error(), "operation was canceled"):
```

provider/postgis/postgis.go
```diff
@@ -201,38 +207,43 @@ func BuildDBConfig(cs string) (*pgxpool.Config, error) {
...
 		// The AfterConnect call runs everytime a new connection is acquired,
 		// including everytime the connection pool expands. The hstore OID
-		// lookup should only be run once
-		if !hasInit {
-			// as extension hstore does not have a fixed OID we have to get it on connect
-			var hstoreOID uint32
+		// is not constant, so we lookup the OID once per provider and store it.
+		// Extensions have to be registered for every new connection.
+
+		if !hstore.hasInit {
```
ARolek, line 217: nice! I like this `hasInit` check.

provider/postgis/postgis.go
```diff
@@ -201,38 +207,43 @@ func BuildDBConfig(cs string) (*pgxpool.Config, error) {
...
 			row := conn.QueryRow(ctx, "SELECT oid FROM pg_type WHERE typname = 'hstore';")
-			if err = row.Scan(&hstoreOID); err != nil {
+			if err = row.Scan(&hstore.OID); err != nil {
 				switch {
 				case strings.Contains(err.Error(), "no rows in result set"):
```
ARolek, line 221: We could probably clean up this check by switching to:

```go
case errors.Is(pgx.ErrNoRows):
```

https://pkg.go.dev/github.com/jackc/pgx/v4#ErrNoRows

provider/postgis/postgis.go
```diff
@@ -201,38 +207,43 @@ func BuildDBConfig(cs string) (*pgxpool.Config, error) {
 		"application_name":              "tegola",
 	}
 
-	var hasInit bool
+	var hstore hstoreOID
```
ARolek, line 210: If we persist this value outside of the AfterConnect call, then it will only be set on the first connection. If I understand you correctly, the hstore registration needs to happen for every connection. If that's the case, then we should move this var to inside the AfterConnect function.

iwpnd: I'm not sure I understand your concern.
The idea is that I only query the database once (for the first connection) for the OID of hstore, and save the OID value per provider in the hstore var. 
Now on the next connection the query is skipped, however the datatype is registered with the OID of the hstore var.
If I were to add a log line into the `if !hstore.hasInit` I'd only see it once, yet registering the datatype happens over and over again.

provider/postgis/postgis.go
iwpnd: ✅

provider/postgis/postgis.go
iwpnd: pat yourself on the back for that, it was your idea 😄

gdey (approved):
LGTM

PR #821, opened 2021-12-29
Build binaries on pull requests

ARolek:
Previously binaries were only built on push so pull requests from
oustside contributors were not building binaires.

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -1,6 +1,7 @@
 name: On release publish
 on: 
   push:
+  pull_request:
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/on_release_publish.yml
```diff
@@ -18,7 +19,7 @@ jobs:
 
     # when it's not a release build use the commit hash for the version tag
     - name: Set tegola version (use commit hash)
-      if: github.event_name == 'push'
+      if: github.event_name != 'release'
       run: echo ${{ github.sha }} | cut -c1-7 > ${{ github.workspace }}/version.txt
```
gdey, line 23: I wonder if we can replace this with a git attribute filter:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#_keyword_expansion

gdey (approved):
LGTM

PR #824, opened 2022-01-27
ci: publish edge image on push or pr to default_branch_ref

iwpnd:
Whenever something is pushed to Tegola default branch, publish an 0day/edge image.

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -22,11 +22,25 @@ jobs:
       if: github.event_name != 'release'
       run: echo ${{ github.sha }} | cut -c1-7 > ${{ github.workspace }}/version.txt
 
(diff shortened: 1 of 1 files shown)


gdey:
This looks fine to me, but what is the reasoning behind it? Just so that we have a build that is on the latest version branch?

Nifty way of getting the current default branch!

iwpnd:
That way I don’t have to fork the current default in my own image, when I for example want to use Redis SSL or the newest pgx. Edge always indicating that it is potentially not stable.

gdey (approved):
LGTM

gdey:
FYI @ARolek is traveling right now, so I assume he will get to this sometime next week.

ARolek (approved):
LGTM! I like this addition. thank you!

PR #827, opened 2022-02-03
fix: gen_version job missing checkout action

iwpnd:
The gen_version job is missing the code checkout. Therefor getting the default branch errors out with

```
fatal: not a git repository (or any of the parent directories): .git
```


ARolek (approved):
details! LGMT.

Issue #831, opened 2022-02-17
log levels and structured logging


iwpnd:
TLDR:
Let's organize logging in Tegola, revisit log levels and classification and push for json structured logs.

Hi there :wave:,

As discussed on Slack I'd love to discuss the issue of logging in Tegola right now. I want to be transparent here and say that I'm pushing for a discussion, because it's a feature that will benefit the company I'm working for. However, I'm trying to make a case why this benefits the Tegola user in general.

I want to briefly cover three issues I see with logging in Tegola right now.
1. chatty logs
2. inability for the user to set log levels
3. text logs vs json logs

In the following I will try to form problem statements for each of the topics and propose action points that I want to put up for discussion.

### 1. Chatty logs
Problem statement:
Under some circumstances tegola can be very very chatty. While not ideal, I don't think this is too much of an issue for smaller applications. Bigger applications with a lot of users however, in my experience, push logs to providers like datadog/cloudwatch etc to aggregate and analyze them. Noisy applications tend to become a cost factor all of a sudden.

One example is this log line:
```
2022-02-17 09:30:09 [INFO] handle_map_layer_zxy.go:100: map (mapname) has no layers, at zoom 6
```

That occurs when the user configures max and min zoom for a layer. 
Is this an information you'd expect to be logged on INFO level?

Action points:
1. agree on a categorization for logs 

Log Level | Importance
-- | --
FATAL | One or more key business functionalities are not working and the whole system doesn‚Äôt fulfill the business functionalities.
ERROR | One or more functionalities are not working, preventing some functionalities from working correctly.
WARN | Unexpected behavior happened inside the application, but it is continuing its work and the key business features are operating as expected.
INFO | An event happened, the event is purely informative and can be ignored during normal operations.
DEBUG | A log level used for events considered to be useful during software debugging when more granular information is needed.
TRACE | A log level describing events showing step by step execution of your code that can be ignored during the standard operation, but may be useful during extended debugging sessions.

([source](https://sematext.com/blog/logging-levels/))

2. use said categorization to revise log statements across tegola, eg. make example above DEBUG level

### 2. inability for the user to set log levels
Problem statement:
Allowing the user to set the desired log level allows to address different user groups and usage scenarios. While development DEBUG level is the preferred choice, INFO or WARN is what makes it to production to reduce information noise and/or alert fatige.

@gdey created a central log package with everything that is needed to to set loglevels for the application. The current init sets defaults it to INFO with no way for the user to change it on start-up or in the config.

[...]

PR #834, opened 2022-02-25
feat: expose SetLogLevel in cli

iwpnd:
Allow user to set a desired log level on startup with `--log-level Level`.
Omitting the flag results in the default INFO

Hey 👋 

I got to one of my proposition from #831. 
As options I went for the log levels that @gdey has already set - all I did is exposing `SetLogLevel` in the root. 

Let me know if you prefer those levels for the flag instead:

```
very-verbose = TRACE
verbose = DEBUG
default = INFO
quiet = WARN
silent = ERROR
```

Where do the docs live actually? 🤔

### DIFF
--- a/cmd/tegola/cmd/root.go
+++ b/cmd/tegola/cmd/root.go
@@ -14,6 +14,7 @@ import (
 )
 
 var (
+	logLevel   string
 	configFile string
 	// parsed config
 	conf config.Config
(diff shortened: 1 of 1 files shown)


cmd/tegola/cmd/root.go
```diff
@@ -22,14 +23,34 @@ var (
...
+	default:
+		return 0, fmt.Errorf("invalid log level use")
+	}
+}
+
 func init() {
 	// root
 	RootCmd.PersistentFlags().StringVar(&configFile, "config", "config.toml",
 		"path or http url to a config file, or \"-\" for stdin")
+	RootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "INFO",
```
ARolek, line 47: we should probably put the various options here in the long description.

ARolek (commented):
This looks fine to me. Since tegola is not using the centralized log package throughout the codebase yet, should we add a note to the flag to say it's only partially adopted at this point?

iwpnd:
understood. I added the options and labelled the flag as a work in progress.

iwpnd:
@gdey since this is actually your work and not mine, are you okay with this?

ARolek:
closed in https://github.com/go-spatial/tegola/commit/e593b4f8518ea2c399834da0048a50261dce9fe5

PR #839, opened 2022-03-10
chore: remove logAndError

iwpnd:
As discussed I'm removing the `logAndError` helper and let control logging for those three occasions in the error block. If you disagree with me on the `Debug` instead of the `Info` log, please feel free to just change it around. :)

### DIFF
--- a/server/handle_map_layer_zxy.go
+++ b/server/handle_map_layer_zxy.go
@@ -95,12 +95,6 @@ func (req *HandleMapLayerZXY) parseURI(r *http.Request) error {
 	return nil
 }
 
-func logAndError(w http.ResponseWriter, code int, format string, vals ...interface{}) {
-	msg := fmt.Sprintf(format, vals...)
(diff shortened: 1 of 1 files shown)


ARolek (approved):
LGTM.

ARolek:
@gdey do you think Debug or Info should be used in these log statements. That's the only real question in this PR.

gdey (approved):
LGTM, Debug seems good, another good one might be TRACE. But I think debug is good.

gdey:
> @gdey do you think Debug or Info should be used in these log statements. That's the only real question in this PR.

I think debug is fine. I don't think these are really `Info`-level messages, as it might be expected to have these happen, but if you are trying to figure out why it's not happening then opening it up to `debug` would be good. The only other level I can think of that might make sense is `TRACE`, but that's supposed to be even more verbose than `debug`.  So, `debug` is good.

PR #840, opened 2022-03-20
chore: add docker-compose local dev env

iwpnd:
This PR adds a local development environment using [docker compose](https://docs.docker.com/compose/) as proposed [here](https://gophers.slack.com/archives/C9LA1F8F8/p1646922370441369) that makes local development a breeze, even if you're unfamiliar with the project.

1. clone the repository
2. `docker compose up`
3. run test suite
```
RUN_POSTGIS_TESTS=yes RUN_REDIS_TESTS=yes PGSSLMODE=disable PGSSLKEY="" PGSSLCERT="" PGSSLROOTCERT="" PGHOST=localhost PGPORT=5432 PGDATABASE=tegola PGUSER=postgres PGPASSWORD=postgres PGUSER_NO_ACCESS=tegola_no_access go test  ./... -v
```
4. `docker compose down`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,45 @@
+# this file is used to bootstrap services to run tegola tests against
+version: "3"
+
+services:
+  redis:
+    image: redis:6.2.6
+    container_name: redis
+    ports:
+      - 6379:6379
+
+  postgis:
+    image: postgis/postgis:12-3.0-alpine
+    container_name: postgis
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres -d postgres"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    environment:
+      POSTGRES_USER: postgres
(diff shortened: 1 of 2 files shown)


gdey (approved):
LGTM, thank you this will make running the full set of tests much easier

iwpnd:
@ARolek thank you for the review

> add a docker-compose.env file with the various env vars that you have in the run command.

This would only allow me to do something like this in the `docker-compose.yml`:

```
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U ${PGUSER} -d ${PGDATABASE}"]
```

It would not load the `.env` environment variables into our local environment so `go test` can use them. For this to work I'd have to refactor the test with a function to load the `docker-compose.env` or something. So the command to run the local test would still need whatever environment variables tegola uses in tests.

> Add the docker-compose commands to the readme

Once the above is solved :)

ARolek (approved):
LGTM!

Issue #774, opened 2021-05-02
Support conninfo string to connect to postgis provider


flacombe:
Dear maintainers,

Currently, postgis provider can only be reached with several arguments like host, port, dbname, etc... while libpq actually offers a more concise way to connect to a postgresql database.
Unless I skip some existing issue, this hasn't been asked for yet.

See 33.1.1 here: https://www.postgresql.org/docs/13/libpq-connect.html

Example:
```
[[providers]]
name = "test_postgis"   # provider name is referenced from map layers
type = "postgis"        # the type of data provider.
uri = "postgres://tegola:tegolapwd@localhost:5432/tegola"
srid = 3857             # The default srid for this provider. If not provided it will be WebMercator (3857)
```
instead of
```
[[providers]]
name = "test_postgis"   # provider name is referenced from map layers
type = "postgis"        # the type of data provider.
host = "localhost"      # postgis database host
port = 5432             # postgis database port
database = "tegola"     # postgis database name
user = "tegola"         # postgis database user
password = "tegolapwd"           # postgis database password
srid = 3857             # The default srid for this provider. If not provided it will be WebMercator (3857)
```

It would be great to add this support in Tegola configuration please :)

Many thanks in advance

ARolek:
I'm all for this support, we just need someone to run with plumbing this trough. We have a couple of related issues: #370 and #125. Would you be interested in working on this feature?

It might make sense to tackle this one along with upgrading the pgx driver so we can squash #748 at the same time.

flacombe:
Thank you for this acknowledgment

Even if I wish to get involved in opensource projects I use, there are two main challenges I have to tackle prior to bring this to Tegola:
* No practice experience of postgresql's libpq API.
* Don't have any experience in Go so far (that's a nice occasion to begin however).

That's barely possible for me to take this time now, dealing with my workload.
Wait and see

ARolek:
No worries! I figured I would throw out the offer in case you were interested in sending in some code. I'm always happy to help.

ARolek:
@flacombe and @wielski This is now in the v0.15.x branch (thanks to @iwpnd!) if you would like to give it a test.

PR #841, opened 2022-03-20
feat: allow connection URI and add additional config parameters

iwpnd:
Hi 👋 

Now with the new pgx it is easier to pass a connection string which was on the wishlist for some issues (eg. #774). If the `uri` is set in the `config.toml` said `uri` is used without compromise as long as the basic information `user`, `password`, `host`, `port` and `database` are set. The `sslmode` is added if not specifically set.

Something on my personal wish list was the addition of  `"max_connection_idle_time"` and `"max_connection_lifetime"` to the config that default to the [pgx default](https://github.com/jackc/pgx/blob/3ce50c079e874c87c4d53eada409738d8504bac6/pgxpool/pool.go#L18) if not specifically set.

### DIFF
--- a/provider/postgis/README.md
+++ b/provider/postgis/README.md
@@ -6,24 +6,60 @@ The PostGIS provider manages querying for tile requests against a Postgres datab
 [[providers]]
 name = "test_postgis"       # provider name is referenced from map layers (required)
 type = "postgis"            # the type of data provider must be "postgis" for this data provider (required)
-host = "localhost"          # PostGIS database host (required)
-port = 5432                 # PostGIS database port (required)
-database = "tegola"         # PostGIS database name (required)
-user = "tegola"             # PostGIS database user (required)
-password = ""               # PostGIS database password (required)
+
+uri = "postgres://tegola:supersecret@localhost:5432/tegola?sslmode=prefer" # PostGIS connection string (required)
+
+host = "localhost"                # PostGIS database host (deprecated)
+port = 5432                       # PostGIS database port (deprecated)
+database = "tegola"               # PostGIS database name (deprecated)
+user = "tegola"                   # PostGIS database user (deprecated)
+password = "supersecret"          # PostGIS database password (deprecated)
+max_connections = 10              # PostGIS max connections (deprecated)
+max_connection_idle_time = "30m"  # PostGIS max connection idle time (deprecated)
+max_connection_lifetime = "1h"    # PostGIS max connection life time (deprecated)
 ```
 
 ### Connection Properties
 
+Establishing a connection via connection string (`uri`) will become the default connection method as of v0.16.0.
+Connecting via host/port/database is flagged for deprecation as of v0.15.0 but will be possible until v0.16.0 still.
+
+- `uri` (string): [Required] PostGIS connection string
 - `name` (string): [Required] provider name is referenced from map layers
 - `type` (string): [Required] the type of data provider. must be "postgis" to use this data provider
-- `host` (string): [Required] PostGIS database host
-- `port` (int): [Required] PostGIS database port (required)
-- `database` (string): [Required] PostGIS database name
-- `user` (string): [Required] PostGIS database user
-- `password` (string): [Required] PostGIS database password
 - `srid` (int): [Optional] The default SRID for the provider. Defaults to WebMercator (3857) but also supports WGS84 (4326)
-- `max_connections` (int): [Optional] The max connections to maintain in the connection pool. Defaults to 100. 0 means no max.
+
+#### Connection string properties
+
+**Example**
+
+```
+# {protocol}://{user}:{password}@{host}:{port}/{database}?{options}=
+postgres://tegola:supersecret@localhost:5432/tegola?sslmode=prefer&pool_max_conns=10
+```
+
+**Options**
+
+- `sslmode`: [Optional] PostGIS SSL mode. Default: "prefer"
+- `pool_max_conns`: [Optional] The max connections to maintain in the connection pool. Defaults to 100. 0 means no max.
+- `pool_max_conn_idle_time`: [Optional] The maximum time an idle connection is kept alive. Defaults to "30m".
+- `max_connection_lifetime` [Optional] The maximum time a connection lives before it is terminated and recreated. Defaults to "1h".
+
+### [DEPRECATED] Connection Properties
+
+- `uri` (string): [Required] PostGIS connection string
+- `name` (string): [Required] provider name is referenced from map layers
+- `type` (string): [Required] the type of data provider. must be "postgis" to use this data provider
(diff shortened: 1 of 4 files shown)


iwpnd:
woops. `url.Values.Has()` was only added with `1.17` - will fix

provider/postgis/postgis.go
```diff
@@ -195,9 +203,144 @@ type hstoreOID struct {
...
+	host, port, err := net.SplitHostPort(uri.Host)
+	if err != nil {
+		return ErrInvalidURI{
+			msg: fmt.Sprintf("%v", err),
+		}
+	}
+
+	if host == "" {
+		return ErrInvalidURI{
+			msg: fmt.Sprintf("address %v:%v: missing host in address", host, port),
```
gdey, line 232: See above about maybe using specialized error objects.

provider/postgis/postgis.go
```diff
@@ -195,9 +203,144 @@ type hstoreOID struct {
...
+		}
+	}
+
+	if uri.Path == "" {
+		return ErrInvalidURI{msg: "missing database"}
```
gdey, line 237: See above about maybe using specialized error objects

provider/postgis/postgis.go
```diff
@@ -195,9 +203,144 @@ type hstoreOID struct {
...
+	uri := DefaultURI
+	uri, err = config.String(ConfigKeyURI, &uri)
+	if err != nil {
+		return &url.URL{}, nil, err
+	}
+
+	// if uri is set in the config, we add sslmode and return early
+	if uri != "" {
+		if err := validateURI(uri); err != nil {
+			return &url.URL{}, nil, err
```
gdey, line 261: Why are you return a new url.URL object if there is an error; instead of just nil?

gdey (changes requested):

iwpnd:
Thank you very much for the detailed review, @gdey. I’ll read up on error handling some more and fix the issues. 🙏🏻

provider/postgis/postgis.go
```diff
@@ -195,9 +203,144 @@ type hstoreOID struct {
...
+// validateURI validates for minimum requirements for a valid postgresql uri
+func validateURI(u string) error {
+	uri, err := url.Parse(u)
+	if err != nil {
+		return ErrInvalidURI{msg: fmt.Sprintf("%v", err)}
+	}
+
+	if uri.Scheme != "postgres" && uri.Scheme != "postgresql" {
+		return ErrInvalidURI{
+			msg: fmt.Sprintf("invalid connection scheme (%v)", uri.Scheme),
```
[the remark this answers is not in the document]
iwpnd: I'm not sure about the specific errors as there are quiet some with the confined scope of the postgis package. But can do if that's what you wanna see.

provider/postgis/error.go
```diff
@@ -37,3 +37,11 @@ type ErrGeomFieldNotFound struct {
 func (e ErrGeomFieldNotFound) Error() string {
 	return fmt.Sprintf("postgis: geom fieldname (%v) not found for layer (%v)", e.GeomFieldName, e.LayerName)
 }
+
+type ErrInvalidURI struct {
+	msg string
+}
+
+func (e ErrInvalidURI) Error() string {
+	return fmt.Sprintf("postgis: invalid uri. reason: %v", e.msg)
```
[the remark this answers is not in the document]
iwpnd: Okay, hope I understand that correctly. So you use `Err` to indicate that the error is the result of another error down the line as opposed to returning the `Msg` in `Error()`.

provider/postgis/postgis.go
```diff
@@ -195,9 +203,144 @@ type hstoreOID struct {
...
+
+	if uri.Path == "" {
+		return ErrInvalidURI{Msg: "missing database"}
+	}
+
+	return nil
+}
+
+// BuildURI creates a database URI from config
+func BuildURI(config dict.Dicter) (*url.URL, *url.Values, error) {
```
ARolek, line 244: I think we should deprecate the old connection method in favor of using the postgresql URI string. We should do this over 2 versions, so this version we should keep backwards compatibility, but the next one we can remove it. That considered, let's add a TODO code comment here explaining that this function can go away after the cutover is complete.

ARolek (commented):
Overall this LGTM! I think we should deprecate the old config strategy in favor of the connection URI string. We should do this over 2 releases, so we should add a warning during the parsing step now.

provider/postgis/postgis.go
iwpnd: ✅

provider/postgis/README.md
```diff
@@ -6,24 +6,60 @@ The PostGIS provider manages querying for tile requests against a Postgres datab
...
+password = "supersecret"          # PostGIS database password (deprecated)
+max_connections = 10              # PostGIS max connections (deprecated)
+max_connection_idle_time = "30m"  # PostGIS max connection idle time (deprecated)
+max_connection_lifetime = "1h"    # PostGIS max connection life time (deprecated)
 ```
 
 ### Connection Properties
 
+Establishing a connection via connection string (`uri`) will become the default connection method as of v0.17.0.
+Connecting via host/port/database is flagged for deprecation as of v0.16.0 but will be possible until v0.17.0 still.
```
ARolek, line 25: minor detail, but change to v0.15.0 and v0.16.0. We still haven't release v0.15.0 (will do so very soon, I'm just waiting for confirmation that we can migrate to go-embed entirely from one of our users that has a build pipeline on go 1.15)

ARolek (approved):
One minot comment, but LGTM! Thank you for the contribution!

ARolek (approved):

durkie:
is there an example way to connect to a local postgresql database via UNIX socket connection with the new `uri=` scheme? I'm trying loads of different permutations but keep getting errors about missing user credentials or missing port.

iwpnd:
https://github.com/iwpnd/tegola-example-bonn/blob/main/config/config-mvt-postgis.toml

Here is an example repo with a working uri.

iwpnd:
I think you have to escape it or use host as Param. Never tried it myself tho.

postgresql:///dbname?host=/var/lib/postgresql

postgresql://%2Fvar%2Flib%2Fpostgresql/dbname

durkie:
Yeah I haven't had much luck with either of those formats: escaping it with %2F results in parse errors, and not escaping it says "missing port". Adding a "port" via something like `/var/lib/postgresql:5432` or `?port=5432` doesn't seem to work

iwpnd:
Im sorry, can't help right now. Tegola is using pgx under the hood, maybe dig a little if that is an issue with pgx.

ARolek:
@durkie thanks for digging into this. Would be great to get some better error handling on these connection params. I would happily accept a PR if you feel so inclined. ;-)

Issue #597, opened 2019-04-09
[config] env vars are not supported in nested config blocks


ARolek:
dict.Dict needs to implement `UnmarshalTOML` in https://github.com/go-spatial/tegola/blob/master/internal/env/types.go.

PR #844, opened 2022-04-26
fix: replace environment variables in webserver headers

iwpnd:
Using environment variables in `config.Webserver.Headers` would not replace environment variables currently. I therefore added an additional step in `config.Parse` that would iterate over the header and resolve the environment variables if possible. I opted to otherwise raise if the environment variable is not present similar to the behaviour we have right now if any other env is missing. This could also just handle the error and use the original value.

I updated the tests accordingly and added some more to test in isolation. 

As mentioned on slack I'm not sure if that is the correct spot to fix it, so your input is greatly appreciated as usual. :)

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -16,14 +16,17 @@ import (
 )
 
 const (
-	ENV_TEST_PORT        = ":8888"
-	ENV_TEST_CENTER_X    = -76.275329586789
-	ENV_TEST_CENTER_Y    = 39.153492567373
-	ENV_TEST_CENTER_Z    = 8.0
-	ENV_TEST_HOST_1      = "cdn"
-	ENV_TEST_HOST_2      = "tegola"
-	ENV_TEST_HOST_3      = "io"
-	ENV_TEST_HOST_CONCAT = ENV_TEST_HOST_1 + "." + ENV_TEST_HOST_2 + "." + ENV_TEST_HOST_3
+	ENV_TEST_PORT                    = ":8888"
+	ENV_TEST_CENTER_X                = -76.275329586789
+	ENV_TEST_CENTER_Y                = 39.153492567373
+	ENV_TEST_CENTER_Z                = 8.0
+	ENV_TEST_HOST_1                  = "cdn"
+	ENV_TEST_HOST_2                  = "tegola"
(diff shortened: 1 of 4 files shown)


ARolek:
Related issue: https://github.com/go-spatial/tegola/issues/597

config/config_test.go
```diff
@@ -198,11 +207,15 @@ func TestParse(t *testing.T) {
...
+				port = "${ENV_TEST_WEBSERVER_PORT}"
+                
+                [webserver.headers]
+                   Cache-Control = "${ENV_TEST_WEBSERVER_HEADER_STRING}"
+				   Test = "Test"
 
 				[[providers]]
 				name = "provider1"
-				type = "postgis"
+				type = "postgis"
```
gdey, line 218: can we remove the extra space

internal/env/parse.go
```diff
@@ -237,3 +237,32 @@ func ParseFloatSlice(val string) ([]float64, error) {
...
+	case map[string]interface{}:
+		for k, v := range val {
+			switch v.(type) {
+			case string:
+				s, err := ParseString(v)
+				if err != nil {
+					return nil, err
+				}
+
+				d[k] = *s
```
gdey, line 258: would it make sense to check if the `v` if of type  `map[string]interface{}` and if so, call ParseDict on it as well?

gdey (commented):
For the most this looks good, there is a question and a minor nit.

internal/env/parse.go
iwpnd: i dont think `v` can be of type `map[string]interface{}` or can it?

ARolek: @iwpnd I'm pretty sure it can be. I don't believe TOML has a max limit of nesting. For the tegola config we don't really have much nesting, but it seems to me the ParseDict() function should likely account for all primitives as well as a `map[string]interface{}` (which would just call ParseDict() again)

iwpnd:
Any outstanding issues that I need to resolve here? @gdey @ARolek

ARolek:
@iwpnd We're just waiting for @gdey to review. I will give him a bump.

gdey (approved):
LGTM

PR #845, opened 2022-04-28
handle gpkg GEOMETRY as unknown geometry

roelarents:
I would like geopakcage [GEOMETRY](https://www.geopackage.org/spec/#geometry_types) type to not result in a startup error , but instead to be handled as an _unknown type_. 

The layer type is used for rendering the capabilities json and styles. In the capabilities this layer will be [set to unknown](https://github.com/go-spatial/tegola/blob/v0.15.x/server/handle_map_capabilities.go#L140). A [style will not be generated](https://github.com/go-spatial/tegola/blob/v0.15.x/server/handle_map_style.go#L138) and a warning will be logged.


gdey:
@roelarents I'm not coming up with a solution I'm happy with, so I want to outline my thinking, and see if we can work through things.

I don't like returning error values as a a geometry value (https://github.com/go-spatial/tegola/pull/845/files#diff-8fa7820d9f23e65780100996d1175d170ed1fa30b8930e52998f7aaf4dd84a0aR265) even if it's only limited to the gpkg provider.

I could see where someone would want to get the error if the mapping geo type is not supported. 
I, almost, think it makes sense to add a switch here (https://github.com/go-spatial/tegola/blob/b1411adcb9e4e42b30657ebdc8bc92f5a40c51d5/provider/gpkg/gpkg_register.go#L194-L198) to check the error type returned and based on a flag set from the config ( https://github.com/go-spatial/tegola/blob/b1411adcb9e4e42b30657ebdc8bc92f5a40c51d5/provider/gpkg/gpkg_register.go#L222) and passed to the function (https://github.com/go-spatial/tegola/blob/b1411adcb9e4e42b30657ebdc8bc92f5a40c51d5/provider/gpkg/gpkg_register.go#L244) we could support both needs. 

But I am unsure if this is the best way forward. I just know that to me passing an error value for a geometry smells. 

thank you,

Gautam.

roelarents:
I understand your objections and I agree. Maybe a new type could be added to go-spatial/geom, called a "MultiGeometry" or "MixedGeometry" or "AnyGeometry" or something.  I also considered adding a type in the `gpkg` package and returning that.

For now, I've made a new try by just returning `nil` (without an error). Which still has the desired effect on capabilities json and styles.

gdey (approved):
This is fine for now...

gdey:
@roelarents if you could please rebase this to master, I will get it merged in and ready for the next release.

roelarents:
@gdey done and thanks

PR #847, opened 2022-05-02
Add a dont_clean option

roelarents:
We have a use case with a very large (85+ GiB, 20+ layers) dataset with small features that only need to be served on one zoom level. "Small" means that the features are usually not bigger than 1 tile in the single zoomlevel that they are served on. The features in the dataset are validated beforehand (self intersecting, etc). Clipping, simplifying and cleaning/validating are not necessary in our case, and disabling them increases performance. Especially cleaning takes a lot of processing time, which takes the response time above a workable threshold for this dataset. (Especially when clipping is disabled, but that's a [known issue](https://github.com/go-spatial/tegola/blob/v0.15.x/atlas/map.go#L290), I believe.)

tl;dr: an option to disable cleaning/making valid

### DIFF
--- a/atlas/layer.go
+++ b/atlas/layer.go
@@ -17,13 +17,13 @@ type Layer struct {
 	// default tags to include when encoding the layer. provider tags take precedence
 	DefaultTags map[string]interface{}
 	GeomType    geom.Geometry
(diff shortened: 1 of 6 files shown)


gdey (approved):
LGTM

PR #848, opened 2022-05-07
feat: structured logs with uber/zap

iwpnd:
Hi 👋 

As discussed in #831 I added the option to use structured logs with uber/zap while falling back to the currently implemented default if `--logger` is not specifically used. Log levels conform to the current implementation with the exception of log level `TRACE` that is not supported by `zap`. I use `Debug` as a fall back here.

Looking forward for your feedback


internal/log/zap.go
```diff
@@ -0,0 +1,70 @@
...
+var zapLogger *ZapLogger
+
+type ZapLogger struct {
+	*zap.SugaredLogger
+}
+
+// SetOutput...
+// To satisfy log.Interface
+func (z *ZapLogger) SetOutput(io.Writer) {
+	return
```
ARolek, line 19: Interesting. Does zap not support changing the output location?

cmd/tegola/cmd/root.go
```diff
@@ -16,13 +16,25 @@ import (
...
+	case log.ZAP:
+		return nil
+	default:
+		return fmt.Errorf("invalid logger %s", logger)
+	}
+}
+
 func getLogLevelFromString(level string) (log.Level, error) {
 	switch level {
 	case "TRACE":
```
ARolek, line 40: Do we have some consts we can key off of for these log levels?

internal/log/zap.go
iwpnd: You can set the output location within the zap config before instantiating the logger. See [here](https://github.com/iwpnd/tegola/blob/a55b7fb9eec551e28befaa6551c2152d1a6e29fc/internal/log/zap.go#L57).

```
zapconfig.OutputPaths = []string{
    "/var/log/myproject/myproject.log",
  }
```

would also be possible here.

cmd/tegola/cmd/root.go
iwpnd: Not sure if I understand your right here. Log levels are numbers ranging from -2 to 3 e.g. `log.TRACE ` equals -2 and we have to derive this from the users input string.
We could add additional consts like `TRACE_STRING` for that switch statement I guess?

iwpnd:
> The main thing I want to think through here is, do we really want to support 2 types of loggers?

While zap is fast, text logs have the edge in terms of performance because json serialization has an overhead. Giving the user the option is only fair as not every users wants to reap the benefits of json logs, or doesn't even care at all.

But ultimately that's your decision to make of course. :)

iwpnd:
Do you want to move forward with this or drop @ARolek @gdey ?

ARolek:
@iwpnd we should move forward with it! I know others have asked about this as well. Let me bump @gdey so he can provide some input as well.

cmd/tegola/cmd/root.go
```diff
@@ -46,6 +58,8 @@ func init() {
 		"path or http url to a config file, or \"-\" for stdin")
 	RootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "INFO",
 		"work in progress - set log level to: TRACE, DEBUG, INFO, WARN or ERROR")
+	RootCmd.PersistentFlags().StringVar(&logger, "logger", "standard",
```
gdey, line 61: minor nit, might want to use the `log.STANDARD` constant here.

internal/log/zap.go
gdey: I think we should drop SetOutput from the Interface{}, we don't have any calls to it, and having the output set during the configuration of the logger makes the most sense.

gdey (commented):
I think we should drop the  SetOutput from the `log.Interface`

PR #850, opened 2022-05-16
fix: deprecation warning when not using uri for provider connection

iwpnd:
Small fix to the deprecation warning when user chooses to use connection parameters in favour of `uri`.


ARolek (approved):
LGTM

PR #861, opened 2022-06-03
test: add a ttools helper for easier local testing

iwpnd:
More than once I stumble upon this during testing. Previously a local test required to setup the local development environment and passing a bunch of environment variables prior to running `go test ./...`. 

```
docker compose up

RUN_POSTGIS_TESTS=yes \ 
RUN_REDIS_TESTS=yes \ 
PGSSLMODE=disable \ 
PGSSLKEY="" \ 
PGSSLCERT="" \ 
PGSSLROOTCERT="" \
PGHOST=localhost \
PGPORT=5432 \
PGDATABASE=tegola \
PGUSER=postgres \
PGPASSWORD=postgres \
PGUSER_NO_ACCESS=tegola_no_access \
go test  ./... -v
```

Now we allow for environment variables to be passed, but also default back to variables we used across tests and pipelines as default, making testing a little more comfy.

```
RUN_POSTGIS_TESTS=yes \ 
RUN_REDIS_TESTS=yes \ 
go test  ./... -v
```

wdyt?

### DIFF
--- a/internal/ttools/getenv.go
+++ b/internal/ttools/getenv.go
@@ -1,0 +1,16 @@
+package ttools
+
+import (
+	"os"
+)
+
+// GetEnvDefault looks up an environment variable, and return a default value
+// when it is not present
+func GetEnvDefault(key, dvalue string) string {
+	v, ok := os.LookupEnv(key)
+	if !ok {
+		return dvalue
+	}
+
+	return v
+}
--- a/internal/ttools/getenv_test.go
+++ b/internal/ttools/getenv_test.go
@@ -1,0 +1,70 @@
+package ttools
+
+import (
+	"os"
+	"testing"
+)
+
+const (
+	ENV_TEST_VARIABLE       = "test"
+	ENV_TEST_EMPTY_VARIABLE = ""
+)
+
+func setEnv() {
+	os.Setenv("ENV_TEST_VARIABLE", ENV_TEST_VARIABLE)
+	os.Setenv("ENV_TEST_EMPTY_VARIABLE", ENV_TEST_EMPTY_VARIABLE)
+}
+
+func unsetEnv() {
+	os.Unsetenv("ENV_TEST_VARIABLE")
+	os.Unsetenv("ENV_TEST_EMPTY_VARIABLE")
+}
+
+func TestGetEnvDefault(t *testing.T) {
+	type tcase struct {
+		key      string
+		dvalue   string
+		expected string
+	}
+
+	setEnv()
+	defer unsetEnv()
+
+	fn := func(tc tcase) func(*testing.T) {
+		return func(t *testing.T) {
+
+			v := GetEnvDefault(tc.key, tc.dvalue)
+
+			if v != tc.expected {
+				t.Errorf("\n\nexpected: %s \ngot: %s \n\n", tc.expected, v)
+			}
+		}
+	}
+
+	tests := map[string]tcase{
+		"should use default value": {
+			key:      "ENV_THAT_DOESNT_EXIST",
+			dvalue:   "DEFAULT_VALUE",
+			expected: "DEFAULT_VALUE",
+		},
+		"should get variable from environment": {
+			key:      "ENV_TEST_VARIABLE",
+			dvalue:   "DEFAULT_VALUE",
+			expected: ENV_TEST_VARIABLE,
+		},
+		"should use default when key is empty string": {
+			key:      "",
+			dvalue:   "DEFAULT_VALUE",
+			expected: "DEFAULT_VALUE",
+		},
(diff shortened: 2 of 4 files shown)


provider/postgis/postgis_internal_test.go
```diff
@@ -20,28 +19,20 @@ const TESTENV = "RUN_POSTGIS_TESTS"
...
-		fmt.Fprintf(os.Stderr, "err parsing PGPORT: '%v' using default port: 5433", err)
-		return 5433
-	}
-	return int(port)
-}
-
 func getConfigFromEnv() map[string]interface{} {
-	port := GetTestPort()
+	// safe to ignore err, as we always return a value that can be parsed to int
+	port, _ := strconv.Atoi(ttools.GetEnvDefault("PGPORT", "5432"))
```
gdey, line 24: This is not true, as the USER could pass in a value that would not parse to an int.  I think it would be better to have a GetEnvDefaultInt function that takes an int default value and returns an int. The other thing we can do; is create a new type  (say EnvironmentVar) and attach a set of functions to it.
e.g.:
```go

type EnvironmentVar string

// AsString will return the value stored in the environment variable if set  otherwise the default
func (env EnvironmentVar) AsString(default string) string {
    val, ok := os.LookupEnv(string(env))
    if !ok {
      return default
    }
    return val
}

// AsInt will return the value stored in the EnvironmentVar as an int, otherwise a default value. If the value stored in the EnvironmentVar can not be cocersed into an int the default value will be returned along will the cocersion error.
func (env EnvironmentVar) AsInt(default int) (int,error) {
    val, ok := os.LookupEnv(string(env))
    if !ok {
      return default, nil
   }
     v, err := strconv.Atoi(val)
    if err != nill  {
       return default, err
    }
    return v, nil
}
```

provider/postgis/util_internal_test.go
```diff
@@ -124,11 +124,12 @@ func TestDecipherFields(t *testing.T) {
...
 	}
 
-	host := os.Getenv("PGHOST")
-	port := os.Getenv("PGPORT")
-	db := os.Getenv("PGDATABASE")
-	user := os.Getenv("PGUSER")
-	password := os.Getenv("PGPASSWORD")
+	host := ttools.GetEnvDefault("PGHOST", "localhost")
+	// safe to ignore err, as we always return a value that can be parsed to int
+	port, _ := strconv.Atoi(ttools.GetEnvDefault("PGPORT", "5432"))
```
gdey, line 129: See the above comment

gdey (commented):
I am all for making development and especially testing simpler and less error-prone. We should not use `os.Getenv` for this, instead, we should use `os.LookupEnv`.  I have a few suggestions in line.

internal/ttools/getenv.go
```diff
@@ -0,0 +1,14 @@
+package ttools
+
+import (
+	"os"
+)
+
+func GetEnvDefault(key, dvalue string) string {
+	v := os.Getenv(key)
```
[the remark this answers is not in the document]
iwpnd: Ah, nice!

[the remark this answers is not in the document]
iwpnd: I added additional tests. Getenv has a similar behaviour covering that case also.

internal/ttools/getenv_test.go
```diff
@@ -49,6 +52,16 @@ func TestGetEnvDefault(t *testing.T) {
...
+		"should get variable when key is empty string": {
+			key:      "",
+			dvalue:   "DEFAULT_VALUE",
+			expected: "DEFAULT_VALUE",
+		},
+		"should get variable from empty string environment": {
+			key:      "ENV_TEST_EMPTY_VARIABLE",
+			dvalue:   "DEFAULT_VALUE",
+			expected: "DEFAULT_VALUE",
+		},
```
gdey, line 64: I think I was not clear enough. The test is proving the exact opposite of what I was saying. I was saying if an environmental variable was set to an empty string (i.e. setting the Password to empty) it should return an empty string and not the default value. 

So, here I would expect the `expected` value to be `""`, and not `"DEFAULT_VALUE"`.

internal/ttools/getenv.go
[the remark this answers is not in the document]
gdey: I don't think I was being clear enough: https://github.com/go-spatial/tegola/pull/861#discussion_r889612018

internal/ttools/getenv_test.go
iwpnd: Ah, now I got you. Thank you for your patience with me here

gdey (approved):
LGTM

ARolek:
@gdey feel free to merge this.

PR #865, opened 2022-06-14
feat: structured logs with uber/zap

iwpnd:
Re-opening this as discussed.

As discussed in #831 I added the option to use structured logs with uber/zap while falling back to the currently implemented default if `--logger` is not specifically used. Log levels conform to the current implementation with the exception of log level `TRACE` that is not supported by zap.
I use Debug as a fall back here.


gdey:
Mergein with a rebase onto master

PR #871, opened 2022-07-04
fix: filter by zoom level for min/max zoom 0

iwpnd:
Prior to this, setting min_zoom=0 and max_zoom=0 would result
in the layer being present across the entire zoom range.

Should fix #870

Issue #870, opened 2022-06-27
Unexpected shapes from z0 layer included in z13 tile


dwoznicki:
With my custom layer setup, I've noticed that some tiles at higher zoom levels contain shapes from my base zoom level 0 table. This leads to some unexpected results, like the one below where I'd expect the entire 13/1314/3169 to be water.

![image](https://user-images.githubusercontent.com/13933454/176036683-b04da618-320b-431e-b53e-b6d24ee41bb7.png)

It looks to me like this is happening for any tile that does not contain other layer data in it (you can see an extra land polygon in tile 13/1315/3170 which also should be pure water).

<details>
<summary>Here's my tegola.toml file.</summary>

```toml
[webserver]
port = ":8004"

[cache]
type = "file"
# basepath = "/nvme-ssd/tegola-cache"
basepath = "/tmp/tegola-cache"

[[providers]]
name = "all"
type = "mvt_postgis"
host = "localhost"
port = "5435"
database = "vector"
user = "ntk"
password = "storesomething"
max_connections = 300

    #--------------------------------------------------------------------------
    # Land areas
    [[providers.layers]]
    name = "land_areas_z0"
    geometry_fieldname = "geometry"
    id_fieldname = "id"
    geometry_type = "polygon"
    sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) AS geometry, id, class, type, area, 'z0' AS t FROM ntk_land_areas_z0 WHERE geometry && !BBOX!"

    [[providers.layers]]
    name = "land_areas_z1"
    geometry_fieldname = "geometry"
    id_fieldname = "id"
    geometry_type = "polygon"
    sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) AS geometry, id, class, type, area, 'z1' AS t FROM ntk_land_areas_z1 WHERE geometry && !BBOX!"

    [[providers.layers]]
    name = "land_areas_z2"
    geometry_fieldname = "geometry"
    id_fieldname = "id"
    geometry_type = "polygon"
    sql = "SELECT ST_AsMVTGeom(geometry, !BBOX!) AS geometry, id, class, type, area, 'z2' AS t FROM ntk_land_areas_z2 WHERE geometry && !BBOX!"

    [[providers.layers]]
    name = "land_areas_z3"
    geometry_fieldname = "geometry"
    id_fieldname = "id"
    geometry_type = "polygon"
[...]

PR #874, opened 2022-07-24
fix: config max zoom 0

iwpnd:
as per #870, setting max_zoom to 0 currently causes the layer to bleed through
over the entire zoom range. Either we fix filtering as per #871
or make sure max_zoom is never 0 as I do in this PR.


gdey (approved):
LGTM

PR #879, opened 2022-08-26
feat: parameter to control tile logging during seeding

### LINKED ISSUE #878: Add option to only log tiles that are slow to render during seeding
When seeding tiles, I'd like the option to control logging output so that I only see tiles which took a while to render. The current behavior is to log the render time for all tiles, which produces a lot of output at higher zoom levels.

I'd propose a `--log-threshold int` flag, where `int` is the amount of time in milliseconds. Tiles that take less than `int` time to render won't be logged.

dwoznicki:
This PR adds a `--log-threshold int` command line argument to the `seed` command to limit which tiles produce logs when rendered. Tiles that render in less than `int` milliseconds will not produce logging output. The default value is 0, at which all tiles rendered will be logged.

My goal here is to reduce noise and get a better sense of which zoom levels actually need to be cached (as opposed to rendered on the fly).

Fixes #878

### DIFF
--- a/cmd/tegola/cmd/cache/seed_purge.go
+++ b/cmd/tegola/cmd/cache/seed_purge.go
@@ -49,6 +49,8 @@ var (
 	cacheBounds string
 	// name of the map
 	cacheMap string
+	// while seeding, on log output for tiles that take longer than this (in milliseconds) to render
+	cacheLogThreshold int64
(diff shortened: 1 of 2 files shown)


gdey (approved):
LGTM thank you for the submission!

PR #891, opened 2022-11-22
feat: add Google Cloud Storage cache adapter

matheusmatos:
This PR implements Google Cloud Storage as a new cache type.

To use it, you need to configure cache as the example below:

```toml
[cache]
# required
type="gcs"
bucket="your_bucket_name"   # Bucket is the name of the GCS bucket to operate on

# optional
basepath="tegola"           # Basepath is a path prefix added to all cache operations inside of the GCS bucket
max_zoom=8                  # MaxZoom determines the max zoom the cache to persist.
```

The credentials (service account and project_id) are handled by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.


ARolek (approved):
LGTM! I will get this merged into `master` once I cut the new version. Thanks for the contribution!

ARolek:
Merged into master! Thank you for the contribution.

kolisko:
@matheusmatos Thanks! Is possible to set up TTL for gcs cache in config file?

PR #892, opened 2022-11-23
README updates and Upgrading Go version

ARolek:


gdey (approved):
LGTM

We may want to setup when to capitalize Tegola and when it should not be.

ARolek:
> We may want to setup when to capitalize Tegola and when it should not be.

My convention is to always lowercase unless a sentence starts with it.

PR #903, opened 2023-01-16
fix: prometheus with multiple providers

### LINKED ISSUE #886: Issue with prometheus with multiple maps
I've run into an issue with prometheus if i have multiple maps defined in my config.toml .
Tegola tries to register one metrics collector for every map defined. 
Can someone take a look at it or give me a hint how to configure prometheus with multiple maps inside one config.toml?

### Actual Behavior
[tegola_error.log](https://github.com/go-spatial/tegola/files/9575146/tegola_error.log)

### Expected Behavior
tegola starts up and exposes "/metrics" url

### Steps to reproduce the issue.
Generate two maps from two different providers in your config.toml + prometheus observer:

### Sanitized config.toml
`	

	[webserver]
	port = ":3001"              

	hostname = "localhost"
	uri_prefix = "/mvt_basemap"

		[webserver.headers]
		Access-Control-Allow-Origin = "*"
		Access-Control-Allow-Methods = "GET, OPTIONS"
	

	[observer]
	type = "prometheus"

	[[providers]]
	name = "basemap_bac"         				
	type = "mvt_postgis"            	
	host = "localhost"          		
	port = 5432                 		
	database = "basemap_bac"        		
	...   		

		[[providers.layers]]
		name = "water_polygons"    
		...
			  
	[[maps]]
	attribution = "OSM Contributers"
	center = [53.3,-6.4, 7.0]
	bounds = [-6.7394, 53.0545, -6.0178, 53.6024]

	name = "basemap_bac"                             

		[[maps.layers]]
		name = "water_polygons"                         

	[[providers]]
	name = "basemap_vos"         				
	type = "mvt_postgis"            	
	host = "localhost"          		
	port = 5432                 		
	...        		
      		

		[[providers.layers]]
		name = "water_polygons"              
		...

	[[maps]]
	attribution = "OSM Contributers"
	 center = [52.3,7.8, 8.0]
	 bounds = [6.9, 51.8, 8.7, 52.7]

	name = "basemap_vos"                              

		[[maps.layers]]
		name = "water_polygons"                         
					 
		provider_layer = "basemap_vos.water_polygons"  
		...

`

iwpnd:
we encountered an issue where tegola would panic if you'd try to register duplicate metrics collectors.
that would happen if you have more than one mvt provider in the same config file.

this fix would append the provider name to the metric name. to ensure backward compatibility the provider name would only be appended if there were more than one.

Would also resolve #886


gdey:
I'm still looking at this. This was my initial idea (as well), but I want to make sure it makes sense.  Also, why do we need both the array of prefixes and the hash map?

```go

var prefixWithProvider bool
{
   var (
       mvtProviderSet = make (map[string]struct{})
      value = struct{}{}
   )
   for _, aMap := range a.maps {
        mvtProviderSet[aMap.mvtProviderName] = value
   }
   prefixWithProvider = len(mvtProviderSet) > 1
}
var prefix = "tegola"
for _, aMap := range a.maps {
     if prefixWithProvider {
        prefix = fmt.Sprintf("tegola_%s",aMap.mvtProviderName())
    }      
    collectors, err := aMap.Collectors(prefix,o.CollectorConfig)
   if err != nil {
       log.Errorf("failed to register collector (%s) for map: %s ignoring", prefix, aMap.Name)
      continue
   }

   if err = o.Register(collectors...); err != nil {
      log.Error("failed to register collectors %+v",err)
     panic(err)
   }
}
```

Would it be okay to break compatibility a bit here? Breaking compatibility will be changing the name of the observer key from `tegola`, to `tegola_mvt_provider_name`; if I'm reading this correctly. If willing, that could be handled in the `postgis` collector method.

I'm still tracing the code to make sure I understand what exactly is going on. Wanted to get my thoughts out as I have already dropped the ball on this once.

iwpnd:
Cool, was not aware that Len() works on a map like that, which is why I'm jumping hoops there. 

You're not breaking compatibility with this change, as previously it was not possible to track more than one provider with prometheus. 

You are correct however, that ideally we'd re-use the same metric but distinguish different providers with a tag, similar to how you distinguish maps right now.

atlas/atlas.go
```diff
@@ -267,9 +268,25 @@ func (a *Atlas) SetObservability(o observability.Interface) {
...
 
-		collectors, err := aMap.Collectors("tegola", o.CollectorConfig)
+	prefixWithProvider := len(mvtProviders) > 1
+
+	for _, aMap := range a.maps {
+		var prefix = "tegola"
+		if prefixWithProvider {
+			prefix = fmt.Sprintf("tegola_%s", aMap.mvtProviderName)
+		}
+		collectors, err := aMap.Collectors(prefix, o.CollectorConfig)
```
gdey, line 289: After, digging deeper. I see what you mean that it would be a large refactor to get the mvt_provider name in, and I'm not sure; if it would work with how we are using labels. (Labels coming from the URL).

I don't think the suffix should be added here; instead, it should go into the Collector for the Map. There is no need to check to see if we only have one map, and only then suffix the mvt provider name. We can prefix the mvt provider name if it exists.

https://github.com/go-spatial/tegola/blob/master/atlas/map.go#L88-L94

Could be:
```go

	if m.mvtProviderName != "" {
		collect, ok := m.mvtProvider.(observability.Observer)
		if !ok {
			return nil, nil
		}
		return collect.Collectors(prefix+"_"+m.mvtProviderName, config)
	}

```

observability/prometheus/http.go
```diff
@@ -64,13 +64,6 @@ func newHttpHandler(registry prometheus.Registerer, prefix string, URLPrefix str
 		[]string{},
 	)
 
-	registry.MustRegister(
-		handler.inFlightGauge,
-		handler.counter,
-		handler.durationSeconds,
-		handler.responseSizeBytes,
-	)
-
```
gdey, line 73: Do you know why these are getting deleted?

gdey (changes requested):
Looking through it, and taking into account your comments; I think the fix can be simpler.

observability/prometheus/http.go
iwpnd: I was reading up on the .MustRegister method and found that promauto would take care of registering given a registry. As this was not the origin of the error I dropped promauto again. But forgot about reverting this. 

Will fix

atlas/atlas.go
iwpnd: Happy to comply here.

I was wondering however. Prometheus should be fine with reusing a metric, rendering a prefix unnecessary. 
Say you have `latency` in your application. Every provider, every route should send metrics to `latency` with additional meta information (provider, path, map, ..). Separating this into meaningful insights is Grafanas job in the end. 

Because on second thought, using a prefix comes with its on set of annoying perks for the user and tegola. 

Do you create lowercase provider name prefix or leave it?
Do you create mutate `my-provider-name` to `my_provider_name`?

Good luck discovering those in grafana when you do not know what we change them to.

Do you suggest we go ahead with the prefix and fix the underlying issue of registering duplicate names in a separate step? If you do not share those concerns, forget what I wrote. :)

gdey: Sorry, for going dark. 
I think we should investigate how to make labels work. That would be the better way.
This really should be a label, the more I thought about it the more I'm convinced of that.
We should avoid doing stop-gap fixes; which I think altering the metric name would be.

iwpnd:
I found a solution using labels instead. I will close this PR and open a new one with a cleaner approach. Thank you anyway for taking the time and going through this with me @gdey 🙏

PR #893, opened 2022-11-30
HANA database provider

mrylov:
Dear Tegola community,

We, the SAP HANA Spatial Team, would like to add HANA support to the Tegola tile server. [SAP HANA](https://www.sap.com/products/hana.html) is an in-memory database with an [OGC-compliant](http://www.opengeospatial.org/resource/products/details/?pid=1303) [spatial engine](https://help.sap.com/viewer/cbbbfc20871e4559abfd45a78ad58c02/2.0.04/en-US/e1c934157bd14021a3b43b5822b2cbe9.html). A free community edition of SAP HANA is available [here](https://www.sap.com/cmp/td/sap-hana-express-edition.html).

### Content 
This pull request includes:
 - Implementation of the provider (provider and mvtprovider interfaces)
 - Provider documentation
 - Unit tests
 - CI integration using a HANA Cloud instance

### Requirements  
The HANA provider requires the open-source library [go-hdb](https://github.com/SAP/go-hdb) that is used to connect to a HANA instance. Note, this driver requires at least go v.1.18 and to change some dependencies in the vendor directory.

### License
The HANA provider is licensed under the [MIT license](https://github.com/go-spatial/tegola/blob/master/LICENSE.md).

### Responsibilities
We promise to keep tracking of new issues/bugs, CI runs, API changes, documentation 
improvements related to the HANA provider and make sure that they are fixed/implemented promptly.
There will be a contact person at SAP responsible for the maintenance.

PR #915, opened 2023-03-23
fix: prometheus with multiple providers

iwpnd:
we encountered an issue where tegola would panic if you'd try to register duplicate metrics collectors.
that would happen if you have more than one mvt provider in the same config file.

this fix makes provider metrics unique by adding the respective provider name to an array of constant labels. Upon collecting metrics on start up we'd pass unique metrics to the registration step and avoid said panic.

resolves: #886

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -8,10 +8,10 @@ import (
 )
 
 var (
-	Version              = "Version not set"
+	Version              = "version not set"
 	GitRevision          = "not set"
 	GitBranch            = "not set"
-	uiVersionDefaultText = "Viewer not build"
+	uiVersionDefaultText = "viewer not build"
 	Tags                 []string
 	Commands             = []string{"tegola"}
 )
--- a/provider/hana/hana.go
+++ b/provider/hana/hana.go
@@ -26,7 +26,12 @@ import (
 const Name = "hana"
 
 type connectionPoolCollector struct {
-	pool                     *sql.DB
+	pool *sql.DB
+
+	// providerName the pool is created for
+	// required to make metrics unique
+	providerName string
+
 	maxConnectionDesc        *prometheus.Desc
 	currentConnectionsDesc   *prometheus.Desc
 	availableConnectionsDesc *prometheus.Desc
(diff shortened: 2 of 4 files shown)


internal/build/build.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 var (
-	Version              = "Version not set"
+	Version              = "version not set"
 	GitRevision          = "not set"
 	GitBranch            = "not set"
 	uiVersionDefaultText = "Viewer not build"
```
gdey, line 14: Since we are fixing the Casing, could we fix it here as well. Thank you. 
"Viewer not built" -> "viewer not built"

gdey (approved):
LGTM

This is so much simpler, and the better solution! Thank you!

internal/build/build.go
iwpnd: missed that one, my bad.

iwpnd:
@ARolek @gdey 

Can I suggest we merge #916 to fix the viewer in CI, then #893 to add the HANA provider and then #915 after I added unique metrics to the HANA provider in #915 as well?

gdey:
> @ARolek @gdey
> 
> Can I suggest we merge #916 to fix the viewer in CI, then #893 to add the HANA provider and then #915 after I added unique metrics to the HANA provider in #915 as well?

I'm good with this. @ARolek can you get this reviewed? The changes seem straight forward, but I'm not well verst in Node.

@iwpnd Looks like there is still there the CI is still failing in building the UI?

ARolek:
> Can I suggest we merge https://github.com/go-spatial/tegola/pull/916 to fix the viewer in CI, then https://github.com/go-spatial/tegola/pull/893 to add the HANA provider and then https://github.com/go-spatial/tegola/pull/915 after I added unique metrics to the HANA provider in https://github.com/go-spatial/tegola/pull/915 as well?

Yes! I just merged in #916. We will land HANA provider next.

provider/postgis/postgis.go
```diff
@@ -34,6 +34,11 @@ const Name = "postgis"
 
 type connectionPoolCollector struct {
 	*pgxpool.Pool
+
+	// provider the pool is created for
```
ARolek, line 38: nit, but the `provider` -> `providerName` in the comment

ARolek (approved):
Wow! That's way cleaner. Thanks for sticking with it!

gdey:
@iwpnd #893 is merged into mainline, when you have a chance, could you rebase, and update the Hana providers as needed?

provider/hana/hana.go
```diff
@@ -387,7 +394,13 @@ func CreateProvider(config dict.Dicter, maps []provider.Map, providerType string
...
+	name, err := config.String(ConfigKeyName, nil)
+	if err != nil {
+		return nil, err
+	}
+
 	p := Provider{
+		name:      name,
 		dbVersion: uint(majorVersion),
 		srid:      uint64(srid),
 		pool:      &connectionPoolCollector{pool: conn},
```
gdey, line 406: Shouldn't you be passing in the ProviderName to the connectionPoolCollector here?

gdey (changes requested):
I think you are forgetting to pass in the provider name to the connection pool for the Hana driver. @mrylov Can you give this a quick review as well? Thank you.

provider/hana/hana.go
iwpnd: good catch, long week 👍

gdey (approved):
LGTM

mrylov:
Thank you @ARolek, @gdey, @iwpnd for the review!

PR #920, opened 2023-03-29
chore: comment hana connection string

iwpnd:
As per @mrylov comment here: https://github.com/go-spatial/tegola/pull/893\#discussion_r1136602372 the connection string is exposed on purpose. Let's comment for future reference.


ARolek (approved):
LGTM

PR #926, opened 2023-05-02
MapLibre - Vue3 - Vite Migration v2

mapl:
Hi everyone, this is my 2nd attempt to migrate the internal UI Viewer project.

All the UI (dev) libraries have been updated to the most recent versions.

Mapbox -> MapLibre
Vue2 -> Vue3
Vue Cli -> Vite

### DIFF
--- a/ui/src/globals/store.js
+++ b/ui/src/globals/store.js
@@ -1,6 +1,6 @@
-import Vue from "vue";
+import { reactive } from "vue";
 
-export const store = Vue.observable({
+export const store = reactive({
   // activeMap is the map the user is currently interacting with
   activeMap: null,
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ tmp
 .DS_Store
 ui/node_modules
 ui/dist
+ui/build
 
 # local env files
 .env.local
--- a/ui/index.html
+++ b/ui/index.html
@@ -1,0 +1,14 @@
+
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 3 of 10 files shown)


ui/src/globals/store.js
```diff
@@ -1,6 +1,7 @@
-import Vue from "vue";
+// import Vue from "vue";
```
iwpnd, line 1: lets clean this up

iwpnd (changes requested):
i checked out the PR, build and tested the viewer and did not find anything out of the ordinary. seems to be working just right.

some nitpicks. i would like to get rid of traces of mapbox in the naming and clean up comments that are still there left an right. but thats about it.

looks good 👍

ARolek (commented):
Looking good in here! Thanks for tackling this. I don't have any additional comments beyond what @iwpnd pointed out.

ARolek (commented):
Looking good in here! Thanks for tackling this. I don't have any additional comments beyond what @iwpnd pointed out.

ui/src/globals/store.js
mapl: thanks for pointing out some obsolete mapbox references. There are still some more left, but these will be a topic for another commit.

ARolek:
@iwpnd it looks like all your comments have been addressed. Would you mark the PR approved if you don't have any additional change requests?

iwpnd (approved):

iwpnd:
my bad, wasn't aware that I was appointed a reviewer and was blocking this

PR #930, opened 2023-06-05
HANA: Set sessionVariable:APPLICATION in connection string

mrylov:


ARolek:
It looks like the Docker build is failing in the CI. I will investigate and try to push the fix to your branch.

PR #938, opened 2023-08-03
fix: cache/gcs dropped errs

alrs:
This picks up two dropped `err` variables in `cache/gcs`.

### DIFF
--- a/cache/gcs/gcs.go
+++ b/cache/gcs/gcs.go
@@ -47,13 +47,21 @@ func New(config dict.Dicter) (cache.Interface, error) {
 	}
 
 	gcsCache.Basepath, err = config.String(ConfigKeyBasepath, nil)
+	if err != nil {
+		return nil, err
+	}
+
 	defaultMaxZoom := uint(tegola.MaxZ)
+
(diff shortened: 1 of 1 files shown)


cache/gcs/gcs.go
```diff
@@ -47,8 +47,16 @@ func New(config dict.Dicter) (cache.Interface, error) {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	defaultMaxZoom := uint(tegola.MaxZ)
+
 	gcsCache.MaxZoom, err = config.Uint(ConfigKeyMaxZoom, &defaultMaxZoom)
+	if err != nil {
+		log.Fatal(err)
+	}
```
ARolek, line 59: please return the error rather than log fatal.

ARolek (changes requested):
Good catch! Please change the error handling to a return.

ARolek:
> Should I change that one as well?

Ooof. Yes please!

alrs:
I've replaced the `log.Fatal()` with a `return`, as well.

ARolek (approved):
LGTM!

PR #941, opened 2023-08-12
Use docker buildx to build multi-platform images

### LINKED ISSUE #940: Push images for more architectures to Docker Hub registry
Hi, my name is Liubomyr and I'm a big fan of your project. I'm writing to you today because I have a suggestion that I think could improve it.

Sadly, currently `gospatial/tegola` only has images for `linux/amd64` architecture, which means that it is not possible to use official image on different architectures like `linux/arm64`.
I think it would be great to add official images for other architectures. The [buildx](https://github.com/docker/buildx) that can be used to build multi-platform images. There is also an official [GitHub action](https://github.com/docker/setup-buildx-action) that can be used as well.

I would be happy to help with this, if needed, to make sure that it is possible to use official image on architectures different from `linux/amd64`.

manastyretskyi:
Closes #940

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -188,7 +188,12 @@ jobs:
       with:
         ui: true
 
-    - name: Build and export to Docker
+    # Buildx can only output single-platform result image to docker
+    # so for testing purposes we are building only for linux/amd64
+    # Later when we publish the image we build for linux/amd64 and linux/arm64
+    # Note that linux/amd64 is cached, so when we build images for publishing
+    # we are not rebuilding linux/amd64
+    - name: Build docker image for testing
       uses: docker/build-push-action@v4
       with:
         context: .
@@ -197,7 +202,7 @@ jobs:
         build-args: ${{ env.BUILD_ARGS }}
         tags: ${{ env.DOCKERHUB_ORG }}/${{ env.DOCKERHUB_REPO }}:${{ env.VERSION }}
 
-    - name: Test Image Build
+    - name: Test image build
       run: |
         docker run --rm ${DOCKERHUB_ORG}/${DOCKERHUB_REPO}:${VERSION} version
 
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -161,42 +161,63 @@ jobs:
     env:
       DOCKERHUB_ORG: gospatial
       DOCKERHUB_REPO: tegola
+      BUILD_ARGS: |
+        BUILDPKG=${BUILD_PKG}
+        VER=${VERSION}
+        BRANCH=${GIT_BRANCH}
+        REVISION=${GIT_REVISION}
 
     steps:
     - name: Check out code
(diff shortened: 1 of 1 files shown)


.github/workflows/on_release_publish.yml
```diff
@@ -161,42 +161,63 @@ jobs:
...
       with:
         ui: true
 
-    - name: Build and tag Docker container
+    - name: Build and export to Docker
+      uses: docker/build-push-action@v4
+      with:
+        context: .
+        load: true
+        platforms: linux/amd64
```
ARolek, line 196: did you intentionally leave out `linux/arm64` here? If so, why?

.github/workflows/on_release_publish.yml
```diff
@@ -161,42 +161,63 @@ jobs:
...
+        build-args: ${{ env.BUILD_ARGS }}
+        tags: ${{ env.DOCKERHUB_ORG }}/${{ env.DOCKERHUB_REPO }}:${{ env.VERSION }}
+
+    - name: Test
```
ARolek, line 200: Please give this step a better name. At first glance I thought it was leftover test code.

ARolek (changes requested):
Looks really good! I just have one minor request and then a clarification. Thanks for the quick turn around.

.github/workflows/on_release_publish.yml
manastyretskyi: Basically, when you build an image with `buildx`, by default, it will not be available for usage, eg, starting actual container with it, so here I'm passing option `load: true` so that it will output image into the docker, but it [can only output single-platform result image](https://docs.docker.com/engine/reference/commandline/buildx_build/#output:~:text=The%20docker%20export%20type%20writes%20the%20single%2Dplatform%20result%20image%20as%20a%20Docker%20image%20specification%20tarball%20on%20the%20client.%20Tarballs%20created%20by%20this%20exporter%20are%20also%20OCI%20compatible.), so for the purpose of testing if image is build properly, I have to do one build here, and later make another builds that will actually be pushed to Docker Hub (At least that is what official [docs](https://docs.docker.com/build/ci/github-actions/test-before-push/) are suggesting)

.github/workflows/on_release_publish.yml
manastyretskyi: Done, I hope new one is better.

.github/workflows/on_release_publish.yml
ARolek: I see. Please add some code comments here so the future us understands what's going on.

manastyretskyi: Done. Hope comment is clear enough.

ARolek: Looks good to me.

ARolek (approved):
LGTM! The CI is failing, but I believe that's due to PRs not have access to secrets. I'm going to merge and give it a run.

Thanks for the contibution!

ARolek:
Looks like the build is successful!

image


I will get a version cut this week so we can use a version tag other than `edge`.

PR #947, opened 2023-08-30
Fix docker image viewer UI

### LINKED ISSUE #945: v0.18.0 Docker image has no viewer UI
As of the changes in #940/#941, the image that is pushed on release loads only the following HTML at /:

```
<pre>
<a href=".keep">.keep</a>
</pre>
```

I'm guessing this is because the "Publish Docker container" step does not include `context: .`, and thus is working on the Git context, which does not include the local filesystem changes like the addition of `ui/dist`.

manastyretskyi:
Fixes #945

Adding `context: .` fixes issue with missing files. I've pushed image to my own [repository](https://hub.docker.com/r/manastyretskyi/tegola) so that you would be able test it as well.

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -210,6 +210,7 @@ jobs:
       if: github.ref == env.DEFAULT_BRANCH_REF
       uses: docker/build-push-action@v4
       with:
+        context: .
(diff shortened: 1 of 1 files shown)


ARolek (approved):
LGTM! Thanks for testing too.

Issue #944, opened 2023-08-29
Map Loading/Unloading On-Demand


jchamberlain:
Hello!

This ticket is trying to do two things: 1) consolidate some details of prior requests for dynamic configuration, 2) specify the requirements of my current project at GeoStrategies. I believe these two can go well together. I'm deliberately (mostly) staying away from implementation details in this first post so we can discuss just the requirements.

Previous/related issues: #57, #58, #262, #613, #933

# Requirements

## 1. Load/Unload Maps On-demand with Zero Downtime
The primary objective is for the app to keep running and all existing maps to continue uninterrupted service while new maps are added and old maps are retired.

Since a map's layers must all exist in the provider, and layers may be created on-the-fly (SQL with various filters/transformations applied), this implies:

#### 1-A. Load/Unload Providers On-demand
We require the ability to create new providers and remove old providers without downtime just as with maps. Critically, a provider's lifecycle must be tied in some way to the map's lifecycle, at the very least guaranteeing that the provider is created before and removed after any dependent maps.
    
#### 1-B. Load New Map in <100ms
This is a rough guideline, but we want no perceivable delay between a user action triggering a map load and that map being available.
    
#### 1-C. Load New Map Without Clearing Existing Connection Pools
Existing maps should continue service uninterrupted and undegraded. This means their providers can't be reinstantiated, i.e., we can't reload the whole configuration.

## 2. Synchronize Map Loading/Unloading Across Cluster
The method by which maps+providers are loaded/unloaded must not hinder horizontal scaling/HA of Tegola instances.

#### 2-A. Use Shared Config
All instances in a cluster should be able to look at the same set of maps+providers. They should not require individual configuration, e.g., connecting to each instance and adding a config file.

#### 2-B. Poll/Watch/Subscribe to Map/Providers Loads/Unloads
All instances in a cluster should be able to fetch or be notified of map+provider changes, keeping requirement I.B in mind (polling >10 times/sec to meet that goal is probably not ideal)

> [!NOTE]
> All of this on-demand config change is limited to maps and providers. I see no need to make the remainder of the TOML file reloadable (at least in this manner). I see a strong distinction between general app/system config (like caching, observability, etc) which applies at boot time and maps which change constantly.

> [!NOTE]
> I also see no need to modify existing maps or providers. For our purposes, we always add a new one and retire the old, with unique names being generated automatically.

PR #948, opened 2023-09-01
POC: watch config files to load/unload maps

jchamberlain:
This is a first draft to demonstrate an approach to solving for #944 (Dynamic loading of maps). This is not meant to be ready for merge, but I've run it locally and it works well so far. Please give it a spin, and I'd love your feedback on the acceptability of the overall approach. Thanks!

## New Abstraction
The central new idea/concept/abstraction is the "App". (I'm sure there's a better term for this, but bear with me.) An App is a collection of providers and maps which share a lifecycle. They are loaded into a Tegola instance as a unit and, when no longer needed, unloaded from Tegola as a unit.

The base TOML configuration file may contain an initial App which will never be unloaded. (Which is to say, the `[[providers]]` and `[[maps]]` section of the TOML file still work exactly as expected.)

## Configuring Tegola to load Apps dynamically
A new, optional section is added to the base config, `[app_config_source]`. It includes a `type` plus any options required for that source type. The only option for the "file" type is `dir` which sets a directory from which to load apps.

```toml
[app_config_source]
type = "file"  # could also be "s3" or "redis" or any other source type someone wants to implement
dir = "./apps" # if relative, this path is relative to the location of the regular config file
```

Add any TOML file to that directory and its `[[providers]]` and `[[maps]]` will be registered immediately. Delete the file, and they'll be removed.

## Key Implementation Details
The `config/source` package has been added providing `ConfigSource` and `ConfigWatcher` (implemented only by `FileConfigSource` so far):

```go
type ConfigSource interface {
	Type() string
	LoadAndWatch(ctx context.Context) (ConfigWatcher, error)
}

type ConfigWatcher struct {
	Updates   chan App
	Deletions chan string
}
```

The `initConfig` process of Tegola's root command has been augmented with `initAppConfigSource()`, which starts a goroutine to receive updates and deletions from the `ConfigWatcher` channels.

To ensure that apps are loaded and unloaded as a unit, `Atlas` has been given two new methods, `AddMaps()` and `RemoveMaps()`. Name collisions are checked before adding any maps to `Atlas`, so that we don't quit partway through having added some maps but not others.


atlas/atlas.go
```diff
@@ -221,6 +222,52 @@ func (a *Atlas) AddMap(m Map) {
...
+		// Use the default Atlas if a, is nil. This way the empty value is
+		// still useful.
+		return defaultAtlas.AddMaps(maps)
+	}
+	a.Lock()
+	defer a.Unlock()
+
+	if a.maps == nil {
+		a.maps = map[string]Map{}
+	}
```
gdey, line 237: since you already know the map is nil here it would be better to:
```go
if a.maps == nil {
   a.maps = make(map[string]Map,len(maps))
   for _, m := range maps {
      a.maps[m.Name] = m
   }
   return nil
}
```

cmd/tegola/cmd/root.go
```diff
@@ -152,3 +156,124 @@ func initConfig(configFile string, cacheRequired bool, logLevel string, logger s
...
+
+	return nil
+}
+
+// initAppConfigSource sets up an additional configuration source for "apps" (groups of providers and maps) to be loaded and unloaded on-the-fly.
+func initAppConfigSource(conf config.Config) error {
+	// Get the config source type. If none, return.
+	val, err := conf.AppConfigSource.String("type", nil)
+	if err != nil || val == "" {
+		return nil
```
gdey, line 190: Why are we not passing up the error here? Did you mean to swallow the error here?

cmd/tegola/cmd/root.go
```diff
@@ -152,3 +156,124 @@ func initConfig(configFile string, cacheRequired bool, logLevel string, logger s
...
+	}
+
+	// Initialize the source.
+	ctx := context.Background() // Not doing anything with context now, but could use it for stopping this goroutine.
```
gdey, line 194: This function should take in a context. Have the calling function pass in context.Background if needed.

cmd/tegola/cmd/root.go
```diff
@@ -152,3 +156,124 @@ func initConfig(configFile string, cacheRequired bool, logLevel string, logger s
...
+		return err
+	}
+
+	// Load and start watching for new apps.
+	watcher, err := src.LoadAndWatch(ctx)
+	if err != nil {
+		return err
+	}
+
+	go func() {
```
gdey, line 206: Export this to another function, and pass in ctx and watcher to that function.

cmd/tegola/cmd/root.go
```diff
@@ -152,3 +156,124 @@ func initConfig(configFile string, cacheRequired bool, logLevel string, logger s
...
+				if app, exists := apps[deleted]; exists {
+					log.Infof("Unloading app %s...", app.Key)
+					register.UnloadMaps(nil, getMapNames(app))
+					delete(apps, app.Key)
+				} else {
+					log.Infof("Received an unload event for app %s, but couldn't find it.", deleted)
+				}
+
+			case <-ctx.Done():
+				return
```
gdey, line 264: We are not closing out database connections, or provider connections here. Are we just leaking those?

config/source/file.go
```diff
@@ -0,0 +1,134 @@
...
+func (s *FileConfigSource) Init(options env.Dict, baseDir string) error {
+	var err error
+	dir, err := options.String("dir", nil)
+	if err != nil {
+		return err
+	}
+
+	// If dir is relative, make it relative to baseDir.
+	if !filepath.IsAbs(dir) {
+		dir = filepath.Join(baseDir, dir)
```
gdey, line 33: We probably should do a filepath.Clean here as well?

gdey (commented):
Not the direction I was initially thinking, but it seems workable. Things I'm worried about:

Leaking resources, especially database connections. I don't see providers being unregistered when shutting down. I've left a few comments. I will need to build this branch and play with it with a sample dataset.

jchamberlain:
>  I don't see providers being unregistered when shutting down....

Good catch. My impression was that providers lived only as pointers on each map, but I didn't realize each driver/provider type is keeping track of its instantiated providers. This causes a couple problems with my approach:

1. There's no master list of all providers, so we can't check for name uniqueness. (This PR is ok with that—names must be unique only within an "app" because only maps in the app can use its providers, so there's no ambiguity.)
2. Without unique names, drivers/provider types don't have a way of unloading individual providers.

atlas/atlas.go
jchamberlain: Good call. Will do.

cmd/tegola/cmd/root.go
jchamberlain: Yes, the swallowing is intentional. `Dict.String()` will return an error if the key is not present and no default is provided. This is an optional field with no default, so `Dict` is going to return errors when nothing erroneous has happened.

cmd/tegola/cmd/root.go
jchamberlain: 👍

config/source/file.go
jchamberlain: `filepath.Join` calls `Clean` internally. https://pkg.go.dev/path/filepath#Join

jchamberlain:
> Actually, if we add a `Cleanup()` method to each provider, we don't need a master list. We can grab the provider from the map(s) as we unload the maps.

Despite my previous comment, I ended up creating a global list of provider instances in `provider/provider.go`. However, I prefer their names with a namespace (in the case of app config files, the file name) so uniqueness is required only per "app".

cmd/tegola/cmd/root.go
jchamberlain: This is now addressed.

cmd/tegola/cmd/root.go
gdey: If that's the case, ignore the error and check val. 
e.g.
```go

// don't care if an optional "type" entry does not exist in the config.
val, _ := conf.AppConfigSource.String("type",nil)
if val == "" {
   return nil
}
```

gdey (commented):
Just a quick review since the code review changes. I will look at this again once the other two PRs ( #949 and #952 ) have been merged.

jchamberlain:
I have not forgotten this. My work has been focused on integrating our other systems with Tegola since Tegola is working well enough already, but I'll have some time later this month to get back to working on Tegola itself. I'll add the experimental guard to this PR as per @ARolek's request.

PR #949, opened 2023-09-06
Reduce precision to 8 places for query tokens (including BBOX) so that tests pass

jchamberlain:
The purpose of this PR is to make tests pass on Apple silicon without breaking tests for x86. The assumption (possibly incorrect) is that 8 places of precision is enough for scale demoninator, pixel width and height, and bounding box.


jchamberlain:
There's more going on here than just precision. I tried getting the `TestGenerateTilesForBounds` test passing without fully understanding it, but now realize it's failing on `arm64` but not `amd64` because overflow of unsigned ints is implementation-specific (see https://github.com/golang/go/issues/52943). We'll need a different solution for that test case.

gdey:
Sounds like there is still an issue here. Were you looking for advice or looking to find a new way to write this test case?

jchamberlain:
This address the first part of #951.

I've removed the changes to `TestGenerateTilesForBounds` as that is handled in #952. Now this contains only precision formatting changes.

gdey (approved):
LGTM, should wait to deploy after the upgrade to geom PR is through.

ARolek:
I just merged this into master but the PR didn't close. ref: https://github.com/go-spatial/tegola/commit/b7ad33bafd713e72254c01786c45dc441b7c3406

PR #950, opened 2023-09-07
Issue 943 tegola lambda al2 support

ARolek:


.github/actions/amazon-linux-build-action/Dockerfile
```diff
@@ -5,8 +5,8 @@ FROM amazonlinux:latest
 RUN yum install -y tar gzip gcc
 
 # install Go
-ENV GOLANG_VERSION 1.20.2
-ENV GOLANG_VERSION_SHA256 4eaea32f59cde4dc635fbc42161031d13e1c780b87097f4b4234cfce671f1768
+ENV GOLANG_VERSION 1.21.0
```
gdey, line 8: should we update this 1.21.1? As it's out?

.github/workflows/on_release_publish.yml
```diff
@@ -140,27 +141,27 @@ jobs:
     - name: Setup env
       uses: ./.github/actions/tegola-setup-env
       with:
-        ui: false
+        ui: true
```
gdey, line 144: Is there a reason we are building the UI now?

gdey (approved):
LGTM, just has a few minor questions.

.github/actions/amazon-linux-build-action/Dockerfile
ARolek: I can do that.

Issue #951, opened 2023-09-11
Pass tests on Apple silicon


jchamberlain:
Tegola's tests don't pass on M1/M2 Macs for two reasons:

1. Many calculated floats are slightly (insignificantly) different, but SQL string comparisons in the PostGIS and Hana providers expect an exact match.
2. `TestGenerateTilesForBounds` fails because of a bug in `go-spatial/geom` (`uint` overflow, resulting in implementation-specific behavior—see https://github.com/golang/go/issues/52943). The `geom` code in question has since been rewritten and does not exhibit the same bug, so upgrading fixes it.

From my perspective, (2) is best fixed by upgrading `go-spatial/geom`. But that introduces its own trouble in that some float calculations are even more different (exacerbating 1), and backward-incompatible changes have been made.

jchamberlain:
With the two PRs merged, tests are now passing on my M1 Mac. Thank you!

PR #952, opened 2023-09-12
Upgrade go-spatial/geom to latest

jchamberlain:
This addresses the second point in #951. The main changes impacting this project are in the `slippy` package, and include:

* Using a grid for calculating extent, rather than relying on `slippy.Tile` directly.
* Removal of SRID-specific extent functions (`Extent3857()`, `Extent4326()`), thus requiring use a another tool for transformations.
* Buggy behavior in new `slippy.RangeFamilyAt()`, thus requiring keeping around the old version of that function.


atlas/map_test.go
```diff
@@ -345,7 +345,7 @@ func TestEncode(t *testing.T) {
 					},
 				},
 			},
-			tile: slippy.NewTile(2, 3, 4),
+			tile: slippy.NewTile(2, 3, 3),
```
jchamberlain, line 348: 4 is not a valid y-value at zoom 2, and the geom upgrade makes this test fail unless a valid coordinate is used.

server/handle_map_layer_zxy.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/go-spatial/geom"
+	"github.com/go-spatial/proj"
```
jchamberlain, line 12: I'm using `go-spatial/proj` because it's pure Go and it allows converting an arbitrary number of points in one pass. I considered `github.com/wroge/wgs84` which is also pure Go and better maintained, but it's a little more annoying to use.

ARolek:
@jchamberlain could you please split this PR into 2 commits: 1 for the vendoring of `geom` (and go.mod updates) and then a second one for the code changes? Makes review a bit easier.

cmd/tegola/cmd/cache/slippy.go
```diff
@@ -0,0 +1,43 @@
...
+
+	leastX := t.X << mag
+	leastY := t.Y << mag
+
+	for x := leastX; x < leastX+delta; x++ {
+		for y := leastY; y < leastY+delta; y++ {
+			err := f(slippy.NewTile(zoom, x, y))
+			if err != nil {
+				return err
+			}
```
gdey, line 38: might as well combine these lines:
```go
if err := f(slippy.NewTile(zoom,x,y)); err != nil {
   return err
}
```

cmd/tegola/cmd/cache/slippy_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	fn := func(tc tcase) func(t *testing.T) {
+		return func(t *testing.T) {
+
+			coordList := make([]coord, 0, len(tc.expected))
+			rangeFamilyAt(tc.tile, tc.zoomAt, func(tile *slippy.Tile) error {
+				z, x, y := tile.ZXY()
+				c := coord{z, x, y}
+
+				coordList = append(coordList, c)
```
gdey, line 38: no need for the extra `c` variable here.

provider/provider.go
```diff
@@ -86,12 +101,22 @@ func NewTile(z, x, y, buf, srid uint) Tile {
...
 }
 
 // BufferedExtent returns a the extent of the tile, with the define buffer
 func (tile *tile_t) BufferedExtent() (ext *geom.Extent, srid uint64) {
-	return tile.Extent3857().ExpandBy(slippy.Pixels2Webs(tile.Z, tile.buffer)), 3857
+	ext, _ = tile.Extent()
+	// WARNING: slippy.PixelsToNative() is misnamed. It doesn't convert the given number of pixels into
+	// anything--the pixels arguments isn't even used (hence set to 0 here). Instead, slippy.PixelsToNative()
+	// return the ratio of pixels to projected units at the given zoom. We multiply its return value by
+	// the pixel count in tile.buffer to get the expected conversion.
```
gdey, line 118: Let's create an issue around this to get the function named correctly.

gdey (approved):
LGTM

There are some minor code cleanups; that we can get to later. And we should make an issue around renaming slippy.PixelsToNative.

cmd/tegola/cmd/cache/slippy.go
jchamberlain: This code is copy-and-pasted without modification from the older version of `go-spatial/geom`. I would recommend fixing `geom` rather than cleaning this up.

cmd/tegola/cmd/cache/slippy_test.go
jchamberlain: See previous comment. I would prioritize deleting this entirely.

provider/provider.go
jchamberlain: I've opened https://github.com/go-spatial/geom/issues/128 and https://github.com/go-spatial/geom/issues/129.

ARolek:
@jchamberlain can you rebase against master? I just fixed the docker hub issue (and upgraded Go in the CI) so we can get the CI to pass. Then I can merge this in.

PR #967, opened 2024-02-06
[FIX] Issue 904 -  Vue app not taken into account uri_prefix 

### LINKED ISSUE #134: Specify client protocol in config
When running behind a reverse proxy, SSL may be available, so HTTPS URLs should be able to be generated automatically.

Nactik:
# Fix uri_prefix not taken into account for the UI

Hey there, here's a quick PR to fix the way tegola UI is served when sitting behind a reverse proxy.

## Changes:
- Added a new setting `proxy_protocol` (I can change it if you think of a better name) in the config file (`config.toml`) to set a custom protocol that will be used to generate the URLs included in the capabilities endpoint responses. (Useful when working under a reverse proxy and when the `X-Forwarded-Proto` is not a solution - Fix #134 )
- When serving embedded files under a reverse proxy, strip the `uri_prefix` from the request (`server/viewer_embed.go`)
- Adjust `vite` settings to include static file using a relative path

## Still to do:
- Add tests - comming soon


ARolek:
@Nactik nice! LMK when this is out of draft and I will give it a review.

Nactik:
@ARolek 
It should be good now :)

server/server.go
```diff
@@ -50,6 +50,8 @@ var (
 	// when the server sits behind a reverse proxy with a prefix (i.e. /tegola)
 	URIPrefix = "/"
 
+	ProxyProtocol string
```
iwpnd, line 53: Reading the PR the option is self-explanatory. Without context it is not. A comment would be nice here - or a ref to this PR at least. 🙏

iwpnd (approved):
Nice, lgtm. 👍 

was not aware `http.StripPrefix` was a thing.

server/server.go
Nactik: Oupsi daisy !
It would indeed be better to explain this param. I've just added the comment :)

config/config_test.go
```diff
@@ -477,6 +479,163 @@ func TestParse(t *testing.T) {
...
+				Providers: []env.Dict{
+					{
+						"name":     "provider1",
+						"type":     "postgis",
+						"host":     "localhost",
+						"port":     int64(5432),
+						"database": "osm_water",
+						"user":     "admin",
+						"password": "",
+						"layers": []map[string]interface{}{
```
ARolek, line 574: nit, but can you change the type `interface{}` to `any`. This was changed a few Go versions ago, so I try to catch it when I can. It's a benign detail

ARolek (approved):
This looks excellent! I left a couple of nits line, but no blockers. Can you please squash your commits into a single commit and then I will get this merged in? 

🙏

config/config_test.go
Nactik: Done :)

ARolek:
Just merged this into master. Great for contribution! Thank you.

PR #978, opened 2024-02-24
ci: Add govulncheck job. Update actions with deprecation warnings

ARolek:
In this PR the following changes are made:

* Add `govulncheck` job. This will no run the go team vulnerability scan on every code change and report any issues.
* Upgrade all deprecated actions: Several actions were throwing deprecation warnings due to them using Node.js 16. I updated all actions that were throwing warnings to the newest versions available.


gdey (approved):
LGTM

PR #985, opened 2024-03-18
Bump Go version to 1.21.8

ARolek:
@gdey this bumps Go up to 1.21.8 to address some of the stdlib vulns. govulncheck is still failing on a few other dependencies that I will update after this one lands.


gdey (approved):
LGTM

PR #986, opened 2024-03-20
refactor: deprecate non uri connection

### LINKED ISSUE #979: provider/postgis: Clean out deprecated connection params
In v0.15.0 support for a connection string uri was added to provider/postgis. When this change was made, we marked the [old connection params as deprecated](https://github.com/go-spatial/tegola/tree/master/provider/postgis#deprecated-connection-properties). These params now need to be removed. This issue is for tracking this removal from the following places:

- [ ] `provider/postgis`:  This package needs to remove support for the old connection params. The README should also be updated.
- [ ] `.github/`: The CI uses the old connection params in a few places and needs to be migrated.

iwpnd:
closes #979

### DIFF
--- a/mvtprovider/postgis/README.md
+++ b/mvtprovider/postgis/README.md
@@ -1,32 +1,48 @@
 # PostGIS MVT Provider
 
-The PostGIS MVT provider manages querying for tile requests against a Postgres database (version 12+) with the [PostGIS](http://postgis.net/)(version 3.0+) extension installed and leverages [ST_AsMVT](https://postgis.net/docs/ST_AsMVT.html) to handle the MVT encoding at the database. 
+The PostGIS MVT provider manages querying for tile requests against a Postgres database (version 12+) with the [PostGIS](http://postgis.net/)(version 3.0+) extension installed and leverages [ST_AsMVT](https://postgis.net/docs/ST_AsMVT.html) to handle the MVT encoding at the database.
 
 The connection between tegola and PostGIS is configured in a `tegola.toml` file. An example minimum connection config:
-
 
 ```toml
 [[providers]]
 name = "test_postgis"       # provider name is referenced from map layers (required)
 type = "mvt_postgis"        # the type of data provider must be "mvt_postgis" for this data provider (required)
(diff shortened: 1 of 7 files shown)


mvtprovider/postgis/README.md
```diff
@@ -1,32 +1,22 @@
...
 ### Connection Properties
 
-- `name` (string): [Required] provider name is referenced from map layers.
-- `type` (string): [Required] the type of data provider. must be "postgis" to use this data provider
-- `host` (string): [Required] PostGIS database host
-- `port` (int): [Required] PostGIS database port (required)
-- `database` (string): [Required] PostGIS database name
-- `user` (string): [Required] PostGIS database user
-- `password` (string): [Required] PostGIS database password
-- `max_connections` (int): [Optional] The max connections to maintain in the connection pool. Defaults to 100. 0 means no max.
```
ARolek, line 28: We should add a reference around some of these pgx specific configs. It looks like the [`ParseConfig` ](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#ParseConfig) function for pgxpool supports the following additional query strings:

* pool_max_conns: integer greater than 0 
* pool_min_conns: integer 0 or greater 
* pool_max_conn_lifetime: duration string 
* pool_max_conn_idle_time: duration string 
* pool_health_check_period: duration string 
* pool_max_conn_lifetime_jitter: duration string

I think adding some notes to the README about how this package uses `pgx` under the hood with a link with to the ParseConfig function and indicating some of the available connection config query strings would be really helpful.

ARolek (changes requested):
Looks really good in here! I only have some minor comments inline, mainly around adding some documentation referencing the pgxpool.ParseConfig function which has additional connection string params that we have deprecated config support for, but are still supported in the connection string.

gdey (approved):
LGTM, when you have a chance can you rebase the commits into a single narrative.

ARolek (approved):
Looks clean! Thanks for tackling that. Please squash the commits into a single one and then we can land this PR and get v0.20.0 of tegola released!

iwpnd:
lets go @ARolek

PR #992, opened 2024-05-23
cache: allow to set a local proxy endpoint for s3

elukey:
Hi folks!

At Wikimedia we use Tegola on Kubernetes and recently we noticed a performance regression when the internal S3 endpoint changed its TLS certificate (more info https://phabricator.wikimedia.org/T344324). We currently use version 0.19.x, we haven't tested the more recent ones, but we'd like to use a local sidecar/proxy to improve HTTP connection pooling and similar settings.

We are currently connecting from the pod directly to the internal S3 endpoint, and we may explore other alternatives if you don't like the one proposed in this patch. The only one that comes to mind is adding explicit settings to tune HTTP connection pooling and similar in tegola's AWS sdk config for S3.

Technical details:
If Tegola is configured to use a s3 cache endpoint like http://localhost:1234, the AWS request signing step is not done correctly and the real S3 endpoint will return a HTTP 403 stating that the request signing header value doesn't match what it expects.
This is a typical use case in k8s, where the HTTP connection go through a local sidecar/proxy. 
In our case, we have an internal S3 endpoint (implemented via Openstack Swift) and we noticed an increase in CPU usage in the pod running tegola when the TLS cert of the S3 endpoint changed (probably due to high requirements for the Cipher suite). We would like to implement in the local proxy connection pooling and other similar performance improvements, without necessarily changing any of the Tegola's code or settings. We'd also get HTTP metrics for free from the sidecar, without the need of instrumenting Tegola.

Let me know your thoughts! Thanks in advance :)

### DIFF
--- a/cache/s3/README.md
+++ b/cache/s3/README.md
@@ -22,7 +22,7 @@ The s3cache config supports the following properties:
 - `cache_control` (string): [Optional] the HTTP cache control header to set on the file when putting the file. defaults to ''.
 - `content_type` (string): [Optional] the http MIME-type set on the file when putting the file. defaults to 'application/vnd.mapbox-vector-tile'.
 - `force_path_style` (bool): [Optional] use path-style addressing instead of virtual hosted-style addressing (i.e. http://s3.amazonaws.com/BUCKET/KEY instead of http://BUCKET.s3.amazonaws.com/KEY)
-
+- `req_signing_host` (string): [Optional] force AWS request signing to use a different Host value, useful when `endpoint` is set to a a local proxy/sidecar.
 
 ## Credential chain
 If the `aws_access_key_id` and `aws_secret_access_key` are not set, then the [credential provider chain](http://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) will be used. The provider chain supports multiple methods for passing credentials, one of which is setting environment variables. For example:
@@ -43,3 +43,12 @@ $ export AWS_REGION=TEST_BUCKET_REGION
 $ export AWS_ACCESS_KEY_ID=YOUR_AKID
 $ export AWS_SECRET_ACCESS_KEY=YOUR_SECRET_KEY
 ```
+
+## Use a local proxy or sidecar (in k8s)
+If `endpoint` is set to a local reverse proxy (like `http://localhost:1234`), then AWS request signing will not work: the real S3 endpoint will return a HTTP 403 error saying:
+
+```
+SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
+```
+
+To make it work, the `req_signing_host` is a special parameter that forces Tegola to use a different HTTP Host header value when the AWS sdk signs the request to be sent to the real S3 endpoint. It needs to be set to the Host header (so no http:// prefixes etc..) of the real S3 endpoint (behind the reverse proxy for example).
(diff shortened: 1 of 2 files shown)


cache/s3/s3.go
```diff
@@ -141,6 +144,19 @@ func New(config dict.Dicter) (cache.Interface, error) {
 		return nil, err
 	}
 
+	// If a Proxy/Sidecar/etc.. is used, the Host header needs to
+	// be fixed to allow Request Signing to work.
+	// More info: https://github.com/aws/aws-sdk-go/issues/1473
+	req_signing_host := DefaultReqSigningHost
+	req_signing_host, err = config.String(ConfigKeyReqSigningHost, &req_signing_host)
```
gdey, line 151: `req_signing_host := DefaultReqSigningHost` this does nothing. As it's going to get immediately overwritten in the next line:
`req_signing_host, err = config.String(ConfigKeyReqSigningHost, &req_signing_host)` Just do:
`req_signing_host, err := config.String(ConfigKeyReqSigningHost, &req_signing_host)`

gdey (changes requested):
Minor Nit, I would like to see fix, but other seems good. I have not tested the code.

cache/s3/s3.go
elukey: Hi! I followed what it is done for all the other parameters, I thought it was a style choice. If I try to build with the line you proposed I get:

```
# github.com/go-spatial/tegola/cache/s3
../../cache/s3/s3.go:149:67: undefined: req_signing_host
```

gdey: Did you change the `=` to an `:=` like I did in my comment? That's why you are getting the error.

gdey: Nevermind, I see what I missed. disregard my whole comment. I for some reason did not see that `req_signing_host` was being passed into config.String as the default value. 

Disregard the entire comment, and revert back to what you had. That is correct.

gdey (approved):
LGTM

elukey:
Hi @ARolek! When you have a moment: what do you think about the PR? It is running in Wikimedia production at the moment, looks working as expected :)

ARolek:
👋 @elukey I will have some time tomorrow to give this a review. Thanks for your submission!

cache/s3/s3.go
```diff
@@ -141,6 +144,19 @@ func New(config dict.Dicter) (cache.Interface, error) {
 		return nil, err
 	}
 
+	// If a Proxy/Sidecar/etc.. is used, the Host header needs to
+	// be fixed to allow Request Signing to work.
+	// More info: https://github.com/aws/aws-sdk-go/issues/1473
+	req_signing_host := DefaultReqSigningHost
```
ARolek, line 150: minor nit here, but can you change the var naming to match the rest of the conventions in this func? For example:

`req_signing_host` -> `reqSigningHost`. Minor detail, but I would like to keep the conventions consistent this close together.

cache/s3/README.md
ARolek, line 1: Thanks for adding the associated documentation.

ARolek (approved):
This looks great! I have one minor nit inline. I approved this PR with the assumption that you will address this nit. Thanks for the submission, I will get this included in the upcoming release.

cache/s3/s3.go
elukey: Definitely! I used Python syntax in a Go file, my brain confused the two :) Hope that now works, lemme know @ARolek!

ARolek: Haha, no worries. Looks good! I'm just waiting for the CI and then I will merge.

Issue #869, opened 2022-06-26
custom provider query -> wrong/too much results


baststar:
Hi,
i have this layer configured:

```
[[providers.layers]]
    name = "admin_lines_level_8"
    geometry_fieldname = "geometry"
    id_fieldname = "osm_id"
    sql = "SELECT ST_AsBinary(geometry) AS geometry, osm_id, admin_level, name, type FROM osm_admin_areas WHERE admin_level = '8' AND geometry && !BBOX!"
```
with this map:

```
[[maps.layers]]
    name = "admin_lines_level_8"
    provider_layer = "osm.admin_lines_level_8"
    min_zoom = 14
    max_zoom = 18
```

and my provider-configuration:

```
    [[providers]]
    name = "osm"                         
    type = "postgis" 
    host = "postgres.default.svc" 
    port = 5432
    database = "osm" 
    user = "tegola"
    password = "xxx"
    srid = 3857
```

but instead of getting data only where admin_level = 8 i get also all other admin_levels (2, 4, 6, 9)

Am i doing this right or am i misunderstanding something?

Im using no cache at the moment

ARolek:
Are the in levels integers or text in the database? Your query has '8' which indicates text. Double check the data type on the admin levels column. 

Also, for the provider, take a look at mvt_postgis as you will see some great performance gains.

iwpnd:
While you seem to be correct that [admin level](https://wiki.openstreetmap.org/wiki/Key:admin_level) is type int, the expected outcome of using `'8'` would be 0 rows, and not all rows. 🤔

baststar:
where is the output from the logs? 
I'm using kubernetes, i think debugging is enabled? but the service doesnt output debug-infos and /var/log is empty in the tegola-pod

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: tegola
spec:
  selector:
    matchLabels:
      app: tegola
      tier: frontend
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app: tegola
        tier: frontend
    spec:
      containers:
        - name: tegola
          #image: gospatial/tegola:v0.14.0
          image: registry.gitlab.com/bamdelicious/tegolamirror:latest
          imagePullPolicy: Always
          ports:
            - containerPort: 8080
          resources:
            limits:
              cpu: 2000m
              memory: 8Gi
            requests:
              cpu: 2000m
              memory: 8Gi
          env:
            - name: TEGOLA_SQL_DEBUG
[...]

PR #995, opened 2024-06-27
Layer names that are substrings cause atlas to serve all layers that match

KoduIsGreat:
refers to #869 

use direct comparison instead of strings.Contains to filter layernames

happy to discuss if we need to support both approaches.

### DIFF
--- a/atlas/map.go
+++ b/atlas/map.go
@@ -167,7 +167,7 @@ func (m Map) FilterLayersByName(names ...string) Map {
 	nameStr := strings.Join(names, ",")
 	for i := range m.Layers {
 		// if we have a name set, use it for the lookup
-		if m.Layers[i].Name != "" && strings.Contains(nameStr, m.Layers[i].Name) {
+		if m.Layers[i].Name != "" && nameStr == m.Layers[i].Name {
 			layers = append(layers, m.Layers[i])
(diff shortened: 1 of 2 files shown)


ARolek (approved):
Looks good! Thanks for hunting down that bug. That has been there for a very long time. 

We will need to detail this change in the release notes just in case people see some changes in the data returned after upgrading. This is nuanced though so I don't suspect it has hit too many people.

gdey:
@ARolek seems our dockerhub login key has expired.

iwpnd:
Nah, this is because the PR does not have access to the credentials I believe. If you were to merge this now, it will have them @gdey.
Maybe we can split the docker build step and make the publish step including login only happen on push to master.

iwpnd (approved):

ARolek:
> seems our dockerhub login key has expired.

This has to do with external PRs not being able to access the repo secrets. I have been trying to figure out how we could make that job conditional based on if you have access to the secrets or not. [Here's a related discussion](https://github.com/actions/runner/issues/1138). Seems like a few work arounds also reference this iss.

ARolek:
@KoduIsGreat can you please squash all the commits into a single one? I can help with this if you would like me to do it.

PR #996, opened 2024-07-01
ci: build docker only when secrets are present

iwpnd:
as discussed in #995 docker build should only run if `secrets.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_TOKEN` are present.

### DIFF
--- a/.github/workflows/on_release_publish.yml
+++ b/.github/workflows/on_release_publish.yml
@@ -1,12 +1,13 @@
 name: On release publish
 on: 
-  push:
   pull_request:
+  push:
+    branches:
+      - master
   release:
     types: [published]
 
 jobs:
-
   gen_version:
     name: Generate software version
     runs-on: ubuntu-22.04
@@ -223,6 +224,8 @@ jobs:
     needs: [gen_version,build_ui]
     runs-on: ubuntu-22.04
     env:
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
       DOCKERHUB_ORG: gospatial
       DOCKERHUB_REPO: tegola
       BUILD_ARGS: |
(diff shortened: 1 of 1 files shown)


ARolek:
@iwpnd thanks for jumping on this! I think you updated the wrong action though. I believe you meant to update https://github.com/go-spatial/tegola/blob/master/.github/workflows/on_pr_push.yml but instead updated the release action.

iwpnd:
The failing docker build is part of `on_release`. If you do not want to build at all in PRs we can also remove the 

```
on:
  pull_request:
```

from the release workflow. In that case we can also skip the check on whether the secrets are defined, because it only runs after the PR is merged. @ARolek

ARolek:
@iwpnd ah yes, you're right! I remember when this was added years ago. It had to do with errors popping up during release that didn't get caught during the PR. So let's leave the check. Your changes LGTM! 🙏

ARolek (approved):

ARolek:
@iwpnd it appears we have a syntax error in the workflow file:

image

iwpnd:
> @iwpnd it appears we have a syntax error in the workflow file:
> 
> image
> Edit: Workflow run: https://github.com/go-spatial/tegola/actions/runs/9751755854

I’ll take a look in the morning. Seems I don’t have access to secrets in the if statement after all. 🤔

sorry for the noise 🙏🏻

ARolek:
> I’ll take a look in the morning. Seems I don’t have access to secrets in the if statement after all. 🤔
> sorry for the noise 🙏🏻

No worries! I was hoping this worked as it was a clean solution.

iwpnd:
okay TIL i dont have access to secrets, but I can export secrets to the environment variables and use that instead. 😄 

@ARolek now when we merge the secrets should be populated in those environment variables and the docker build should execute successfully and not be skipped.

ARolek:
> okay TIL i dont have access to secrets, but I can export secrets to the environment variables and use that instead.

Interesting! Still looks clean to me ;-)

ARolek (approved):
LGTM! Can you please squash into a single commit?

Issue #1010, opened 2024-11-17
add support for username for redis as cache in tegola config


ibrahimjel:
Hi,
I am using tegola with redis as cache type. my redis DB has username and password to authenticate with. when put in address like this URL `redis://<username>:<password>@redis:6379`, tegola throw exception 

> could not register cache: address  `redis://<user>:<password>@redis:6379`: too many colons in address

would be great to introduce new field `username` in [redis.go](https://github.com/go-spatial/tegola/blob/master/cache/redis/redis.go#L63).

iwpnd:
I can take a look tomorrow. I’m however surprised to read this. Pretty sure we use it with credentials too.

iwpnd:
You can use the uri like you did. Make sure your password does not contain a colon though. If this does not work correctly I can take a look.

ARolek:
@ibrahimjel here's what seems like a related issue in the driver: https://github.com/redis/go-redis/issues/2870

Try removing `redis://` from the connection string.

iwpnd:
As far as I can see we do not support the uri after all. Probably something we should do similar to the change we did to Postgres a while back? @ARolek 
I mean deprecating the host name in v0.21.0 and removing it in favor of the uri in v0.22.0?

ARolek:
@ibrahimjel the one and only @iwpnd implemented connection string handling for redis! You can grab a binary to test it out via the GH Action: https://github.com/go-spatial/tegola/actions/runs/11917492487. Config documentation has also been updated: https://github.com/go-spatial/tegola/tree/master/cache/redis

Please report back how this works for you.

iwpnd:
`Username` is the same as `REDIS_USER_TEGOLA`. Make sure your environment is properly populated and [your redis instance actually requires a password](https://github.com/redis/redis/issues/13437).

PR #1011, opened 2024-11-18
feat: allow connection to redis using uri

iwpnd:
closes: #1010


iwpnd:
I would upgrade redis in a separate PR. v9 passes contexts to commands, so some changes are required there. Will otherwise address your comments momentarily. :)

ARolek (approved):
Looks good!

PR #1023, opened 2024-12-20
Enhance server hostName handling

ARolek:
The hostName func is used to dynamically determine the hostName value in various API endpoints. The previous logic favored the Host value of request URL, but this value is often not set (i.e. when not operating behind a proxy). The default handling has been changed to return the request Host value if the request URL Host is not set.

This PR also addresses the outstanding ui vulns.


iwpnd (approved):
👀👍

gdey (approved):
LGTM

PR #1024, opened 2024-12-24
When generating the seed list, we need the grid for the bounds

gdey:
Our old system normalized the bounds to grid, this meant that people could provide a bounds without the associated srid as long as it was either 4326 or 3853. With the change over, we were defaulting to 3853. Added option to allow for the grid to be 4326.


iwpnd (approved):
yessss, that looks better! 🚀

ARolek (commented):
@gdey thanks for digging into this! You found the problem, but I'm wondering if the default behavior is the opposite of what we want. Up until this bug surfaced the bounds was always expected to be in 4326. With this change we now require all the users to add `--wgs84=true` when seeding. It seems we should default to wgs84. Thoughts?

gdey:
> @gdey thanks for digging into this! You found the problem, but I'm wondering if the default behavior is the opposite of what we want. Up until this bug surfaced the bounds was always expected to be in 4326. With this change we now require all the users to add `--wgs84=true` when seeding. It seems we should default to wgs84. Thoughts? 

If 4326 was the default then i agree we should swap it. 
I did not realize that when i first fix the bug, because we our math was slightly off and we had a bunch of small adjustments. This came from me not fully understanding things at the time i built it.

gdey:
After discussion we decided it would be better to provide a flag to state the SRID of the bounds.

```

ARolek (approved):
LGTM!

PR #1032, opened 2025-03-05
chore: bump pgx to v5

iwpnd:
# description

- bump pgx to v5
- refactor registering types on connect
- hstore now is a buildin type with changed behaviour, it was required to treat it separately in `decipherFields`


provider/postgis/postgis.go
```diff
@@ -126,7 +127,10 @@ type Provider struct {
 	queryHistogramSeconds            *prometheus.HistogramVec
 }
 
-func (p *Provider) Collectors(prefix string, cfgFn func(configKey string) map[string]interface{}) ([]observability.Collector, error) {
+func (p *Provider) Collectors(
+	prefix string,
+	cfgFn func(configKey string) map[string]interface{},
```
ARolek, line 132: Let's change `interface{}` -> `any`

provider/postgis/postgis.go
```diff
@@ -315,44 +314,33 @@ func BuildDBConfig(opts *DBConfigOptions) (*pgxpool.Config, error) {
 		return nil, err
 	}
 
-	dbconfig.ConnConfig.LogLevel = pgx.LogLevelWarn
```
ARolek, line 318: Note to self. Has this log level config been moved to somewhere else?

provider/postgis/postgis.go
```diff
@@ -315,44 +314,33 @@ func BuildDBConfig(opts *DBConfigOptions) (*pgxpool.Config, error) {
...
 	dbconfig.ConnConfig.RuntimeParams = opts.GetRuntimeParams()
-	var hstore hstoreOID
+
+	var hstore uint32
```
ARolek, line 319: maybe we rename this var to `hstoreOID`?

provider/postgis/util.go
```diff
@@ -244,7 +243,7 @@ func transformVal(valType pgtype.OID, val interface{}) (interface{}, error) {
 }
 
 // decipherFields is responsible for processing the SQL result set, decoding geometries, ids and feature tags.
-func decipherFields(ctx context.Context, geomFieldname, idFieldname string, descriptions []pgproto3.FieldDescription, values []interface{}) (gid uint64, geom []byte, tags map[string]interface{}, err error) {
+func decipherFields(ctx context.Context, geomFieldname, idFieldname string, descriptions []pgconn.FieldDescription, values []interface{}) (gid uint64, geom []byte, tags map[string]interface{}, err error) {
```
ARolek, line 246: let's change `interface{}` -> `any`

provider/postgis/util.go
```diff
@@ -292,14 +291,32 @@ func decipherFields(ctx context.Context, geomFieldname, idFieldname string, desc
 						tags[k] = v.String
 					}
 				}
+			case pgtype.Hstore:
+				for k, v := range vex {
+					// we need to check if the key already exists. if it does, then don't overwrite it
```
ARolek, line 296: should we log a warning here? seems like this could be helpful debugging info.

provider/postgis/util.go
```diff
@@ -292,14 +291,32 @@ func decipherFields(ctx context.Context, geomFieldname, idFieldname string, desc
...
+					if _, ok := tags[k]; !ok {
+						tags[k] = *v
```
ARolek, line 298: nil check on `v`

provider/postgis/util.go
```diff
@@ -292,14 +291,32 @@ func decipherFields(ctx context.Context, geomFieldname, idFieldname string, desc
...
+					}
+				}
 			case pgtype.Numeric:
 				var num float64
-				vex.AssignTo(&num)
+				vex.Scan(&num)
```
ARolek, line 303: does this not return an error?

ARolek (commented):
Looking good so far! Thank you for kicking off this upgrade. I left some minor comments inline on my first pass.

provider/postgis/postgis.go
iwpnd: > Note to self. Has this log level config been moved to somewhere else?

yes it did. instead of a logger within pgx, logging is extracted to be in the pgx tracelog package. I'll have to take a look on how to make that work for us later on.

iwpnd: Hm, i'd have to build a log adapter for the tracelog package of pgx for our custom logger. 🤔 

Something like so:

```go
        // this would be an instance of slog logger
	logAdapter := log.NewLoggerAdapter(logger)
	tracer := &tracelog.TraceLog{
		Logger:   logAdapter,
		LogLevel: tracelog.LogLevelNone,
	}
	if <pg debug enabled> {
		tracer.LogLevel = tracelog.LogLevelDebug
	}
	conf.ConnConfig.Tracer = tracer
```

there's readily available adapters for slog already. I don't think it make sense to build a custom logadapter for our custom logger if we agree that slog should be come the default logger and we drop str logger and zap. wdyt?

ARolek: I agree that we should move towards adopting `slog`, but I'm not sure if we should land this PR without a logger in place. One option is that we could shift over to the slog refactor, land that, and then circle back to this PR. If we don't want to do that, I think implementing the custom log adapter would be the way to go, even if it's temporary.

iwpnd: There is a logger, tegola logger remains untouched by this. PGX however does not come with logger any longer, but expects the user to pass it if required.
I will add a very simple adapter that reflects the previous behaviour of:

```
dbconfig.ConnConfig.LogLevel = pgx.LogLevelWarn
```

Which is currently logging anything that pgx considers as warning.

iwpnd:
I wanted to test the hstore tags and noticed that both in main and in my branch the tags do not show properly in the ui. Default tags are ignored as well. @ARolek

docker-compose.yml
```diff
@@ -32,6 +32,8 @@ services:
     environment:
       PGUSER: postgres
       PGPASSWORD: postgres
+    volumes:
+      - ./scripts:/scripts
```
iwpnd, line 36: this allows us to extend test tables a little more flexible than downloading a dump, wdyt?

provider/postgis/util.go
```diff
@@ -347,3 +384,25 @@ func ctxErr(ctx context.Context, err error) error {
...
+// the method does nothing.
+func (l *LoggerAdapter) Log(
+	ctx context.Context, level tracelog.LogLevel,
+	msg string, data map[string]any,
+) {
+	if level != tracelog.LogLevelWarn {
+		return
+	}
+
+	log.Warnf("Postgis: %s, %#v", msg, data)
```
iwpnd, line 405: I'm having trouble testing this. Can you think of an action where pgx would emit a warning before? @ARolek

provider/postgis/postgis.go
```diff
@@ -315,45 +316,38 @@ func BuildDBConfig(opts *DBConfigOptions) (*pgxpool.Config, error) {
...
+			Codec: pgtype.HstoreCodec{},
+		})
 
 		// register UUID type, see https://github.com/jackc/pgx/wiki/UUID-Support
-		conn.ConnInfo().RegisterDataType(pgtype.DataType{
-			Value: &gofrs.UUID{},
-			Name:  "uuid",
-			OID:   pgtype.UUIDOID,
-		})
+		pgxuuid.Register(conn.TypeMap())
```
iwpnd, line 350: I was hoping to avoid by using the builtin pgtype codec for uuid, but I just couldn't get it to work without this external dependency.

docker-compose.yml
gdey: I like it, but how do keep these scripts updated?

provider/postgis/postgis.go
gdey: I know you marked this as resolved, but i'm still seeing `map[string]interface{}`?

provider/postgis/postgis.go
```diff
@@ -315,45 +316,38 @@ func BuildDBConfig(opts *DBConfigOptions) (*pgxpool.Config, error) {
...
+	dbconfig.ConnConfig.Tracer = tracer
+
+	var hstoreOID uint32
 
 	dbconfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
-		// The AfterConnect call runs everytime a new connection is acquired,
-		// including everytime the connection pool expands. The hstore OID
-		// is not constant, so we lookup the OID once per provider and store it.
-		// Extensions have to be registered for every new connection.
-
```
gdey, line 327: Has this behavior changed? Is it okay to re scan the `hstoreOID` on every connect now?

gdey (commented):
I have a few questions. Overall it looks good.

docker-compose.yml
iwpnd: That's the good part. Other than the dump file we're currently using, we can actually update and extend those sql without uploading another dump to the repository.

But, now that I look at it again, this name is confusing. It should be `testdata` (is the convention isnt it?).

provider/postgis/postgis.go
iwpnd: Good catch. Must have slipped through during testing. While it's not impacting functionality to leave it like this, we can keep the previous behaviour and reduce overhead of requesting the oid on every connection entering the pool.

provider/postgis/postgis.go
iwpnd: fixed

docker-compose.yml
iwpnd: fixed

iwpnd:
@ARolek can you please give this another look? 👀

docker-compose.yml
```diff
@@ -32,14 +32,17 @@ services:
...
       PGUSER: postgres
       PGPASSWORD: postgres
-    command: 
+    volumes:
+      - ./testdata:/testdata
+    command:
       - /bin/bash
       - -c
       - |
         wget "https://github.com/go-spatial/tegola-testdata/raw/master/tegola.dump"
```
ARolek, line 41: Note, I'm fine adding this file to our repo. It's only 3.5mb. Younger me cared about not having binaries in the repo, but this is test data and I think it should live next to the source code that uses it.

provider/postgis/util.go
```diff
@@ -201,7 +201,7 @@ func uppercaseTokens(str string) string {
 	return provider.ParameterTokenRegexp.ReplaceAllStringFunc(str, strings.ToUpper)
 }
 
-func transformVal(valType pgtype.OID, val interface{}) (interface{}, error) {
+func transformVal(valType uint32, val any) (interface{}, error) {
```
ARolek, line 204: Nit, convert the return `interface{}` to `any`

provider/postgis/util.go
```diff
@@ -278,28 +283,60 @@ func decipherFields(ctx context.Context, geomFieldname, idFieldname string, desc
 					return 0, nil, nil, err
 				}
 				idParsed = true
-				break
+				// NOTE: if it can also be a tag, then breaking here
+				// will never add it to tags
+				// break
```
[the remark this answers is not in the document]
ARolek: @iwpnd I'm not sure honestly. Seems like building some test cases here would be helpful. Maybe leave the `break` statement here with a TODO: comment until we have a chance to write up the test case and prove the fail?

provider/postgis/util.go
ARolek: @iwpnd Looking at the postgres [documentation for errors](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html#PLPGSQL-STATEMENTS-RAISE) it seems like you could need to trigger the call `RAISE WARNING`. One idea (and a possible test case candidate) is to create a database function that would raise the warning:

```sql
CREATE OR REPLACE FUNCTION test_warning_log()
RETURNS void AS $$
BEGIN
  RAISE WARNING 'This is a test warning message';
END;
$$ LANGUAGE plpgsql;
```

And then call it:

```sql
SELECT test_warning_log();
```

It would really be great to have proper integration tests in this package against postgis. Maybe we use something like [testcontainers](https://golang.testcontainers.org/modules/postgres/) to help with this?

ARolek (approved):
This looks excellent! My comments are rather minor, and not blockers. I'm going to approve this PR as is. If you choose to land the slog PR and then adjust LogAdapter in this PR, please re-tag me for review. I would like to see how that all comes together.

ARolek:
@iwpnd review in! Sorry for the delayed response. Heading out on vacation this weekend and it has been an intense week trying to wrap everything up. This PR looks great. Thank you for tackling this upgrade!

docker-compose.yml
iwpnd: Do you want me to pull that pg dump into tegola?

provider/postgis/util.go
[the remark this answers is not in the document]
iwpnd: I'm actually testing this in `util_internal_test.go` ([here](https://github.com/iwpnd/tegola/blob/chore%2Fbump-pgx-v5/provider/postgis/util_internal_test.go#L216)) and the test fails when I add the break statement again.

provider/postgis/util.go
iwpnd: I spend some more time on this. So tracelog is not there to forward PostGIS notifications, but to forward pgx internals. Funny enough the old implementation worked the same. In order to forward PostGIS logs (not pgx logs), you'd have to define a `NotificationHandler` on the `ConnConfig` - something we never supported.

For the time being until we have the slog in here, I forward everything from pgx when our logger is set to `debug`.

docker-compose.yml
iwpnd: I'm pulling the `tegola.dump` into the tegola repository.

iwpnd:
Are you also okay with it @gdey? Feel free to merge if you are. :)

gdey (approved):
LGTM

Issue #1039, opened 2025-03-16
Use slog as default and only logger


iwpnd:
Heya,

as discussed in #1032 and [slack](https://gophers.slack.com/archives/C9LA1F8F8/p1740214680414199) I would like to tackle the refactor from text based standard logger and zap structured logger to a single logger around golangs slog with the goal to simplify logging in Tegola, and moving logging closer to the standard libraries.

## Status quo

- The old implementation uses global variables (e.g., logger, level, IsError, etc.) and manual flag management for log level checks
- even though we log structured logs, we do not use structured logging, we put a message in a structured log, that's it.
 
This approach was necessary 5 years ago, but fell out of time.

### Stage 1
- drop `zap` and std logger in favour of `slog`
- add a `tracelog` to every `Error` log for easier debugging. We achieve this with a custom log handler that adds a stacktrace field and populates it. The custom handler will be passed to `slog.SetDefault()` and will become the global logger.
- use `slog` in `log.Info`,`log.Infof` (...) to keep compatibility and reduce the changes in stage 1 to a minimum
- follow `slog`s approach of `Info`, `Debug`, `Warn` and `Error` to keep things simple
- Drop `Fatal(f)`. `Fatal(f)` is really only an `Error(f)` with an `os.Exit(1)`. The `os.Exit(1)` should be explicit, not implicitly hidden in a fatal log. 
- Drop `Trace`, it's not being used anyways
- logs will go to `os.Stderr` instead of `os.Stdout`, use one output for simplicity.

### Stage 2
- drop `log.Debug, log.Infof` and the likes
- modules will either import and use `slog.Default()`, that at this point is out custom logger, 
- or they receive the logger via dependency injection. 
- Module-specific loggers can be derived by adding context using `With` or `WithGroup`.

Can we agree on this? :)

gdey:
I' 100% for this! Thank you!

ARolek:
@iwpnd I also support this! And thank you for putting momentum behind this update. 

> even though we log structured logs, we do not use structured logging, we put a message in a structured log, that's it.

I agree with this point. One problem I find when implementing structured logging is that the "keys" are littered throughout the codebase in random places. An idea as we go through this transition:

* Let's add a `logging.go` file to each package that implements structured logging.
* In the `logging.go` file we define `const` for each log "key". For example:

```go
package foo

const (
  logKeyRequestID = "request_id"
  logKeyRequestMapName = "map_name"
)
```

[...]

PR #1040, opened 2025-03-16
refactor: drop zap and stdlogger in favour of slog

iwpnd:
As outlined in #1039, I'm starting the move to `slog` in this draft.


cmd/tegola/cmd/root.go
```diff
@@ -89,23 +60,31 @@ func rootCmdValidatePersistent(cmd *cobra.Command, _ []string) (err error) {
...
+		// TODO: enable once we switch to slog.Default
+		// instead of internal/log methods
+		AddSource: false,
+	})
+
+	// Wrap the base handler with our custom handler to add stack traces for errors.
+	handler := log.NewHandler(baseHandler)
+	logger := slog.New(handler).
+		WithGroup("tegola").
+		With("version", build.Version, "pid", os.Getpid())
```
ARolek, line 84: we could add in the git hash and branch here as well: 

https://github.com/go-spatial/tegola/blob/9bac264ffa8d15b80423584c2480b0a811351e27/internal/build/build.go#L11-L13

cmd/tegola/cmd/root.go
```diff
@@ -89,23 +60,31 @@ func rootCmdValidatePersistent(cmd *cobra.Command, _ []string) (err error) {
...
-
-	// set log level before the first log is called
-	level, err := getLogLevelFromString(logLevel)
-	if err != nil {
-		return err
-	}
-	log.SetLogLevel(level)
+func initConfig(configFile string, cacheRequired bool, logLevel string) (err error) {
+	// Parse the provided log level; default to INFO if parsing fails.
+	lvl := log.ParseLogLevel(logLevel)
```
ARolek, line 69: Note, `TRACE` doesn't exist in `slog` so the `log-level` usage needs to be updated.

cmd/tegola/cmd/root.go
```diff
@@ -89,23 +60,31 @@ func rootCmdValidatePersistent(cmd *cobra.Command, _ []string) (err error) {
...
+
+	// Create a base text handler that outputs to stderr.
+	// The AddSource option includes file and line info in each log record.
+	baseHandler := slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		Level: lvl,
+		// TODO: enable once we switch to slog.Default
+		// instead of internal/log methods
+		AddSource: false,
+	})
```
ARolek, line 78: Consider moving this instantiation to the `internal/log` package. You're using it here and inside your tests so it might be better to centralize the code.

internal/log/log_test.go
```diff
@@ -2,365 +2,56 @@ package log
...
-	"regexp"
-	"strings"
+	"context"
+	"log/slog"
 	"testing"
 
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
-	"go.uber.org/zap/zaptest/observer"
+	"github.com/stretchr/testify/require"
```
ARolek, line 9: I haven't used this package before. Is it worth adding as a dev dependency?

ARolek (changes requested):
Looks really good in here! I have a few minor suggestions / nits / questions. I suspect you want to land this ahead of the `pgx` changes so we can get the proper pgx tracing in?

internal/log/log_test.go
```diff
@@ -2,365 +2,56 @@ package log
...
+				logger.Log(context.Background(), tc.level, tc.message)
 			}
 
-			if tc.tst() {
-				t.Errorf("%v , expected level to be set correctly, got %v", name, level)
+			output := buf.String()
+			if tc.expectStack {
+				require.Contains(t, output, "stack", "Error log should include a stack trace")
+			} else {
+				require.NotContains(t, output, "stack", "Non-error log should not include a stack trace")
```
gdey, line 53: Why not use `strings.Contains` or a regular expression:

```
stackRegx := regex.MustCompile(`\w+stack\w+`)
stackRegx.Match(output) 

```

gdey (commented):
I agree with the comments that @ARolek left. I added one about the Contains usage?

internal/log/log_test.go
iwpnd: Force of habit really. I enjoy it. I don’t have to worry about building test helpers in every package over again. But I understand if we want to stay vanilla. 👍

internal/log/log_test.go
iwpnd: Force of habit really. There’s lots of good test helpers in testify and it’s simple and super readable. Larger codebases, with different people working on them tend to be very heterogeneous when it comes to testing. But I’m alright if you both want to remain on std libraries. 👍

cmd/tegola/cmd/root.go
iwpnd: Into an `init()`?

iwpnd:
> Looks really good in here! I have a few minor suggestions / nits / questions. I suspect you want to land this ahead of the `pgx` changes so we can get the proper pgx tracing in?

Either is fine. The adapter will be a different commit one way or another. The old logger has its own adapter for the time being.

cmd/tegola/cmd/root.go
iwpnd: Ill add the revision, but not the branch. That's likely to be master anyways unless you're on a fork or test in a pr.

internal/log/log_test.go
iwpnd: changed and dropped testify. still think testify makes it more readable. :)

iwpnd:
Quick note on testify because it got mentioned in two comments. Testify IMO makes test more readable and verbose than the standard testing library. There's a bunch of helper functions that work for both `assert` and `require`.


```go
package main

import (
    "testing"
)

func TestCalculate(t *testing.T) {
    if Calculate(2) != 4 {
        t.Error("Expected 2 + 2 to equal 4")
    }
}
```

```go
package main

import (
    "testing"
    "github.com/stretchr/testify/assert"
)

func TestCalculate(t *testing.T) {
    assert.Equal(t, Calculate(2), 4)
}

```

There's nice way of handling expected and unexpected errors e.g.:

```go
func TestDoSomething(t *testing.T) {
    // This call should not fail
    err := doSomething(false)
    require.NoError(t, err, "Expected no error, but got an error")

    // This call should produce an error
    err = doSomething(true)
    require.NoError(t, err, "Expected no error, but got an error") 
    // The test will fail immediately here if 'err' is non-nil.
}
```

It also supports mocking, test suits and preparation/cleanup helpers and much more. I'd be happy to tackle the refactor to testify every time I'm working on something - if you guys are okay with it. I'm also completely fine not using it, but I would like you to consider it for a second. :)

internal/log/log_test.go
ARolek: @iwpnd I'm not opposed to testify. I was more asking if you thought the dependency was worth it. ;-)

ARolek:
@iwpnd 

> It also supports mocking, test suits and preparation/cleanup helpers and much more. I'd be happy to tackle the refactor to testify every time I'm working on something - if you guys are okay with it. I'm also completely fine not using it, but I would like you to consider it for a second. :)

I don't generally reach for testify, but if you enjoy it that much I'm not opposed to it. This is a dev dependency, so I don't push back as much. I'm a big fan of having great tooling, but I also like to keep the abstractions to a minimum.

ARolek (approved):
This looks great to me! Thank you for tackling this!

gdey:
I have looked at testify, and have not found it to be useful. I find it to be harder to read. Responses inline.

> Quick note on testify because it got mentioned in two comments. Testify IMO makes test more readable and verbose than the standard testing library. There's a bunch of helper functions that work for both `assert` and `require`.
> 
> ```go
> package main
> 
> import (
>     "testing"
> )
> 
> func TestCalculate(t *testing.T) {
>     if Calculate(2) != 4 {
>         t.Error("Expected 2 + 2 to equal 4")
>     }
> }
> ```
> 

Just a quick note:
The message here should be:
`t.Error("2+2, expected 4, got %d", Calculate(2))`

> ```go
> package main
> 
> import (
>     "testing"
>     "github.com/stretchr/testify/assert"
> )
> 
> func TestCalculate(t *testing.T) {
>     assert.Equal(t, Calculate(2), 4)
> }
> ```

I find this harder to read. Because, I don't know what the log statement is. Also, it's hard to grep for if you are just searching. When a test fails, outside of when you are directly writing it. This can occur when code that depends on your code fails. Also, `asserts` place a special role in my mental model, of being sanity check that are in development builds; that should panic. Not in test code. 

With the simpler standard go code:
1. it's just the equality operator; in the simple cases, and in more complex cases, you should write a function to do equality checks anyway. 
2. The output (following the format of "<thing being tested, expected <expected value>, got <got value>") is easier to find and grep for.  

Now you might say, grepping isn't needed after—all the file and  location of the failed test is printed for you. Yes, true, except when you need to find the test failure in a large code based, that has moved because of large changes, between dev and prod. Then grepping for the text is easier, then having to figure out how the code has mutated to find the new location. Something just running the code works find it. All to this is a very minor argument.

> 
> There's nice way of handling expected and unexpected errors e.g.:
> 
> ```go
> func TestDoSomething(t *testing.T) {
>     // This call should not fail
>     err := doSomething(false)
>     require.NoError(t, err, "Expected no error, but got an error")
> 
>     // This call should produce an error
>     err = doSomething(true)
>     require.NoError(t, err, "Expected no error, but got an error") 
>     // The test will fail immediately here if 'err' is non-nil.
> }
> ```
> 
> It also supports mocking, test suits and preparation/cleanup helpers and much more. I'd be happy to tackle the refactor to testify every time I'm working on something - if you guys are okay with it. I'm also completely fine not using it, but I would like you to consider it for a second. :)

The place where it might be useful, is in the setup and cleanup helpers (main reason I did look at it early on.) But with 1.14, go has been adding helpers (like [Cleanup](https://pkg.go.dev/testing#T.Cleanup). And with go 1.24 we have [context](https://pkg.go.dev/testing#T.Context), in the tests now as well.

I know I'm being the old man yelling about the youngling's and their new fangle-toys. I'm not really seeing the benefit of adding this. 

In my opinion, it's harder to read, and adds a dependency we don't really need as we are not making use of any features that would allow it to add value. I'm always up to changing my mind later.

iwpnd:
Totally fair @gdey. 🙂

internal/log/log_test.go
```diff
@@ -52,9 +51,13 @@ func TestStackTraceHandler(t *testing.T) {
...
-				require.Contains(t, output, "stack", "Error log should include a stack trace")
+				if !strings.Contains(output, "stack") {
+					t.Fatal("expect stack in log")
+				}
 			} else {
-				require.NotContains(t, output, "stack", "Non-error log should not include a stack trace")
+				if strings.Contains(output, "stack") {
+					t.Fatal("did not expect stack to be in log")
+				}
 			}
```
gdey: couldn't this be rewritten to:

```go
if hasStack := strings.Contains(output, "stack"); tc.expectStack != strings.Contains(output, "stack") {
     t.Fatal("stack in log, expected %b, got %b", tc.expectStack, hasStack)
}
```

internal/log/log_test.go
```diff
@@ -1,366 +1,71 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			baseHandler := slog.NewTextHandler(&buf, &slog.HandlerOptions{
+				Level:     slog.LevelDebug,
+				AddSource: true,
+			})
+			handler := log.NewHandler(baseHandler)
+			logger := slog.New(handler).WithGroup("testGroup").With("foo", "bar")
```
gdey, line 35: I'm sorry I just noticed this. 
This function breaks our convention on how test should be written: (I thought I had it documented somewhere, but I can not find it. Will have to write up a document.)

```

func Test{{Name}}(t *testing.T) {

     type tcase struct {
       {{ fields for the testcase }}
     }

     fn := func(tc tcase)(func(*testing.T){
   // or ver(b)
  // fn := func(tc tcase) (string, func(*testing.T){ // this sig is hardly used
     return func(t *testing.T){
       {{ test body here }}
    }})

     tests := map[string]tcase{
       "test_1" : { ...  },
         .
         .
         .
      }

     for name, tc := range tests {
         t.Run(name, fn(tc))
     }
```

I would like to keep all our tests in the same structure.

}

gdey (changes requested):
The test structure is wrong. Sorry, I missed that earlier.

gdey (approved):
LGTM, thank you.

gdey:
@iwpnd I figure you will probably will do a rebase and resolve the conflict. But just in case, I wanted to call your attention to it.

ARolek:
Thanks for tackling this @iwpnd!

PR #1054, opened 2025-09-11
refactor: use maps.Copy for cleaner map handling

fengyuchuanshen:
There is a [new function](https://pkg.go.dev/maps@go1.21.1#Copy) added in the go1.21 standard library, which can make the code more concise and easy to read.


gdey (approved):
LGMT

ARolek (approved):

PR #1069, opened 2026-01-12
feat: add lazy ttl on file cache

### LINKED ISSUE #1068: Feature Request: Add TTL support to file cache backend
Hi,

While working with Tegola’s caching, I noticed that TTL configuration is currently only supported for the Redis cache backend, but not for the file cache.

When using the file cache, cached tiles persist indefinitely unless:

- the Tegola pod is restarted, or
- tiles are manually invalidated

This becomes problematic when underlying data changes infrequently but does change occasionally, those updates effectively never bust the cache in a running system.

I initially assumed I had missed an undocumented ttl parameter for the file cache, but after checking the docs and code, this doesn’t seem to exist.

### Current behavior

- Redis cache supports TTL via backend configuration.
- File cache has no TTL or automatic expiration.
- Cache invalidation for file cache requires manual intervention.

### Expected / Desired behavior

Support an optional TTL for the file cache backend, similar in spirit to Redis, but implemented internally by Tegola.

iwpnd:
Closes #1068 

This introduces a lazy tile cache invalidation for the file tile cache. On request we look up the tile on the FS. If `now + ttl > mtime` we delete the old file and report a miss to the caller.
This approach is simple. The caveat is that unrequested tiles remain on the FS until requested.

cc @ARolek @Edefritz

### DIFF
--- a/cache/file/README.md
+++ b/cache/file/README.md
@@ -6,6 +6,7 @@ filecache uses a file system for caching tiles. To use it, add the following min
 [cache]
 type="file"
 basepath="/tmp/tegola-cache"
+ttl=1
 ```
 
 ## Properties
@@ -13,3 +14,4 @@ The filecache config supports the following properties:
 
 - `basepath` (string): [Required] a location on the file system to write the cached tiles to.
 - `max_zoom` (int): [Optional] the max zoom the cache should cache to. After this zoom, Set() calls will return before doing work.
+- `ttl` (int): [Optional] time to live in seconds for cached tiles. Defaults to 0 (never expires). TTL is evaluated lazily on Get() operations - expired tiles are deleted when accessed but may remain on disk if never requested.
--- a/cache/file/file.go
+++ b/cache/file/file.go
@@ -6,31 +6,34 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/go-spatial/tegola"
 	"github.com/go-spatial/tegola/cache"
 	"github.com/go-spatial/tegola/dict"
 )
 
-var (
-	ErrMissingBasepath = errors.New("filecache: missing required param 'basepath'")
-)
+var ErrMissingBasepath = errors.New("filecache: missing required param 'basepath'")
 
 const CacheType = "file"
 
 const (
 	ConfigKeyBasepath = "basepath"
 	ConfigKeyMaxZoom  = "max_zoom"
+	ConfigKeyTTL      = "ttl"
 )
 
+var defaultTTL = 0
+
 func init() {
-	cache.Register(CacheType, New)
+	cache.Register(CacheType, New) //nolint:errcheck
 }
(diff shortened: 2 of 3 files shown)


cache/file/README.md
```diff
@@ -6,10 +6,12 @@ filecache uses a file system for caching tiles. To use it, add the following min
...
 ```
 
 ## Properties
 The filecache config supports the following properties:
 
 - `basepath` (string): [Required] a location on the file system to write the cached tiles to.
-- `max_zoom` (int): [Optional] the max zoom the cache should cache to. After this zoom, Set() calls will return before doing work.
\ No newline at end of file
+- `max_zoom` (int): [Optional] the max zoom the cache should cache to. After this zoom, Set() calls will return before doing work.
+- `ttl` (int): [Optional] time to live in seconds for cached tiles. Defaults to 0 (never expires). TTL is evaluated lazily on Get() operations - expired tiles are deleted when accessed but may remain on disk if never requested.
```
ARolek: "expired tiles are deleted when accessed but may remain on disk if never requested."

This reads a bit weird to me. Maybe missing the word "not" as in "expired tiles are [not] deleted..."?

cache/file/file.go
```diff
@@ -65,7 +74,13 @@ type Cache struct {
 	// MaxZoom determines the max zoom the cache to persist. Beyond this
 	// zoom, cache Set() calls will be ignored. This is useful if the cache
 	// should not be leveraged for higher zooms when data changes often.
-	MaxZoom uint
+	MaxZoom    uint
+	Expiration time.Duration
```
ARolek: maybe copy the documentation of this config here so it shows up in godoc.

cache/file/file.go
```diff
@@ -83,7 +98,22 @@ func (fc *Cache) Get(ctx context.Context, key *cache.Key) ([]byte, bool, error)
...
+		s, err := f.Stat()
+		if err != nil {
+			return nil, false, err
+		}
+
+		if isExpired(s.ModTime(), fc.Expiration) {
+			pErr := fc.Purge(ctx, key)
+			if pErr != nil {
+				return nil, false, err
+			}
```
ARolek: For this part, do we want to purge on Get? I was under the impression that you were just going to return a "miss" but leave the file on disk and let the subsequent "Set" operation be responsible for overwriting. I'm not sure I'm a fan of a Get (read op) being potentially destructive.

cache/file/README.md
iwpnd: Reads alright for me, but I am not a native. Feel free to suggest the change here, so we can commit straight away.

ARolek:
@iwpnd can you clean up that merge commit on this PR? I can also do it against your branch for you. Just want to get this one landed.

iwpnd:
@ARolek sorry, I am out sick the last two days. Will get back tonight tomorrow morning.

PR #1079, opened 2026-02-23
refactor: postgis connection with environment taking precedence over uri

### LINKED ISSUE #1016: Do not required hardcoded postgres connection uri in config file
Tegola is using PGX to connect to  Postgis enabled postgres database. By default, PGX handles standard libpq env variables like PGHOST, PGDATABASE, PGPORT, PGUSER, PGPASSWORD. However Tegola forces either hardcoding the connection url into the config file or manually setting interpolated env vars. This is clunky and unintuitive (especially if you have a password that contains a character that needs escaping) and could be solved by making the connection uri in the config file optional and allowing pgx to handle validating the connection. directly. 

If multiple postgis providers are needed, postgres has a built in convention for this using the [pgservicefile](https://www.postgresql.org/docs/current/libpq-pgservice.html) which is also supported by default by PGX without any additional configuration needed

an optional `service` parameter could be added to allow switching between multiple providers. This way Tegola does not need to reinvent the wheel and manually handle what PGX can already do.

### LINKED ISSUE #1058: Connection to Postgres erroneously requires password
See [postgis.go](https://github.com/go-spatial/tegola/blob/master/provider/postgis/postgis.go#L234) where a new error state is created if the connection URI does not include a user. A user isn't required [in libpq connection URIs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS). This makes sense if using a [Trust Authentication](https://www.postgresql.org/docs/current/auth-trust.html) with [unix domain sockets](https://www.postgresql.org/docs/current/runtime-config-connection.html)

Given #1044 and #1016 it would be much better, I think, to avoid parsing and attempting to validate the connection parameters (env vars and connection string) and just catch errors from pgx when connecting.

iwpnd:
This PR refactors the PostGIS connection handling to make configuration precedence explicit. In previous issues people - rightfully so - complained that we're too strict to enforce a `uri` in a specific form when there's dozen perfectly valid parameter combinations that can be valid connection string. We do however now enforce strict validation in `env` mode. To make this refactor easier, I isolate responsibilities into dedicated components, rather than having a huge function doing it all.

### env mode vs uri mode

```
if env trigger variables present:
    mode = env
    ignore uri, but enrich with config values
else:
    mode = uri, either uri controls it all, or we can enrich with config values
```

One caveat - with #915 we support multiple providers of the same type. If there's two postgis providers in the same map, environment variables that trigger env mode would be used in both providers. Those cases need to be explicitly handled via uri - at least for the basic connection values.

### Changes

- replaced host/port/database/user/password config with a single uri field. That was long overdue. 
- introduced explicit connection modes: `env` or `uri`
- environment variables take full precedence over URI when any trigger (PGHOST,PGPASSWORD,PGUSER,PGDATABASE,PGPORT) is present. 
  - closes #1016
- removed URI and environment merging; only one mode is active
- added strict validation in env mode (host, port, database, user required).
- loosend validation in uri mode. we let pgx handle the validation of malformed uris instead.
  - closes #1058 
- introduced `ErrEnvIncomplete` with clear diagnostics
- centralized connection logic behind a connector
- split connection flow into selector, planner, builder, validator, connector. expandable to other providers. 
- added explicit connect-time `Ping` with bounded timeout to validate a connection
- wrapped ping failures with resolved connection metadata to make errors verbose
- moved `pool_*` configuration to provider config, provider config takes precendence over uri query runtime parameter to allow pool configuration in env mode
- removed legacy connection-building functions
- improved testability through component isolation

### DIFF
--- a/provider/postgis/postgis.go
+++ b/provider/postgis/postgis.go
@@ -2,13 +2,7 @@ package postgis
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
-	"errors"
 	"fmt"
-	"net"
-	"net/url"
-	"os"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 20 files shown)


provider/postgis/postgis.go
```diff
@@ -879,305 +564,312 @@ func (p Provider) inspectLayerGeomType(pname string, l *Layer, maps []provider.M
...
+//   - id_fieldname (string): [Optional] feature ID column.
+//
+//   - fields ([]string): [Optional] additional fields to include if sql
+//     is not defined.
+//
+//   - srid (int): [Optional] layer SRID (3857 or 4326).
+//
+//   - sql (string): [Required if tablename is not defined] custom SQL
+//     statement. The SQL must include required tokens where applicable.
+func CreateProvider(
```
ARolek, line 627: Maybe we accept a context here so we can pass it to c.Connect().

provider/postgis/conn_builder.go
```diff
@@ -0,0 +1,228 @@
...
+// applyTLS applys the connection plan TLS configuration to the pool config.
+func (b *defaultBuilder) applyTLS(cfg *pgxpool.Config, plan connPlan) error {
+	switch plan.SSLMode {
+	case SSLModeDisable, SSLModeEmpty:
+		cfg.ConnConfig.TLSConfig = nil
+		return nil
+
+	case SSLModeAllow, SSLModePrefer, SSLModeRequire:
+		cfg.ConnConfig.TLSConfig = &tls.Config{
+			InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
gdey: This does not apply here.

provider/postgis/conn_builder.go
```diff
@@ -0,0 +1,228 @@
...
+		// ServerName is required for hostname verification.
+		cfg.ConnConfig.TLSConfig = &tls.Config{
+			ServerName: cfg.ConnConfig.Host,
+		}
+
+	case SSLModeVerifyCA:
+		// Go doesn't support "verify chain but skip hostname" directly.
+		// We must set InsecureSkipVerify=true and perform chain verification ourselves.
+		cfg.ConnConfig.TLSConfig = &tls.Config{
+			InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
gdey: This does not apply here.

gdey (approved):
Thank you for doing this.

LGTM

ARolek (approved):
Thank you for doing this! This is an epic refactor, and it's beautiful work. I only had one minor comment inline. Otherwise let's land this!

provider/postgis/postgis.go
iwpnd: Where would you expect this to origin? CreateProvider is called on init() with no way to pass a context.

ARolek: @iwpnd I just reviewed the call stack in more detail and this is a startup routine not a request path. Whenever I see context.Background() I almost always flag it as something to investigate. I don't think there's anything to do this in case without revisiting our registration pattern, which I don't think is necessary. 

So in summary, nothing to do. ;-)

ARolek:
@iwpnd I have fixed the govulncheck errors in `master`. If you want to rebase, we can make the CI happy. Also, do you want to keep all the commits in this PR?

I'm ready to merge whenever you are. Thanks again!

iwpnd:
@ARolek I'm still not 100p happy with the structure and implementation yet. The functionality is fine tho. :thinking: 
I will be squash merging into a single commit.

ARolek:
@iwpnd I know that feeling. Is there a specific piece that you're unhappy with? I can give a deeper design review if you point me to where you're unhappy.

iwpnd:
@ARolek the connector as orchestrator works I guess. While working on the pmtiles provider I was briefly thinking about a structure that can be used more generally - something that would enforce a certain pattern for all new providers and older providers alike. I suppose we already enforce this contract somewhat loosely with the `provider.Register` and `provider.MVTRegister` so maybe I'm just overthinking this. :) I will however consolidate the files a little, as is with the `conn_` prefix it looks rather messy.

ARolek:
@iwpnd 

> While working on the pmtiles provider I was briefly thinking about a structure that can be used more generally - something that would enforce a certain pattern for all new providers and older providers alike. I suppose we already enforce this contract somewhat loosely with the provider.Register and provider.MVTRegister so maybe I'm just overthinking this

I'm a fan of the registration pattern, but I'm also open to improving on the pattern if you have some ideas. That was implemented so long ago, so I'm sure we can make some improvements. 

What about implementing the pmtiles registration was giving you friction?

>I will however consolidate the files a little, as is with the conn_ prefix it looks rather messy.

👍

iwpnd:
> What about implementing the pmtiles registration was giving you friction?

Sorry, missed that question. If memory serves then Tegola is very married with the concept of layers. I got around it eventually, but it wasn't nice. Pmtiles is a source that just serves an already merged `mvt` tile. There's no layers, just a map.

ARolek:
> if memory serves then Tegola is very married with the concept of layers.

Originally this was true, but once we introduced the MVTProvider this is no longer the case. I figured the pmtiler would just return already formed MVT like we do with the PostGIS provider:

https://github.com/go-spatial/tegola/blob/5b2e82cbae4ef520d3135c8c1cd4c38cde757886/provider/mvt_provider.go#L9-L14

iwpnd:
lets merge this @ARolek? it is a good working base now that is plenty adaptable imo.

ARolek:
@iwpnd yes! I'm traveling right now, back on Monday and will get some of this backlog addressed.

ARolek:
@iwpnd I'm going to merge this into master even though govulncheck is erroring out. I'm going to address those errors in some follow up PRs. Thanks again for tackling this!

